PR #14, opened 2017-11-29
Magpie API 0.2.6 & UI updates

fmigneault:
# Adjusts messages and Improves error handling in UI

## [PAV-416](https://www.crim.ca/jira/browse/PAV-416)
- Internal Server Error and Bad Request HTTP now have the same format as other responses (json)
- magpie/version also have have standard HTTP return code/details
- strip out information not actually requested and already available in matchdict for following requests
keep details only in case of exception/error to help debugging
    * `{bird}/magpie/users/{user}/services`
(BAD  - extra userid & resource types)
    * `{bird}/magpie/users/{user}/resources`
(BAD  - extra username & resource types)
    * `{bird}/magpie/users/{user}/resources/{resource_id}/permissions`
(BAD - extra resource_id & user_name)
    * `{bird}/magpie/users/{user}/services/{service}/permissions`
(BAD - whole service structure returned + username + userid)
- resolve incorrect filtering of services from : `{bird}/magpie/services/types/{type}`
(all types/services returned instead of requested ones) 

## [PAV-474](https://www.crim.ca/jira/browse/PAV-474)
- Group creation returns conflict instead of internal server error
- Message corresponding displayed in UI for conflict / invalid name
- Same process is applied for user creation, for each of the fields

## [PAV-476](https://www.crim.ca/jira/browse/PAV-476)
- refuse user name >64 characters (avoid error on long usernames while keeping token functionality)


tomLandry:
Nice.

Issue #27, opened 2018-04-04
Installation bug


huard:
python setup.py install fails at line
```
from magpie.__meta__ import __version__, __author__, __email__
```
because magpie and its dependencies are not yet installed when first running it.

```
Traceback (most recent call last):
  File "setup.py", line 10, in <module>
    from magpie.__meta__ import __version__, __author__, __email__
  File "/home/david/src/Magpie/magpie/__init__.py", line 5, in <module>
    from api_except import *
  File "/home/david/src/Magpie/magpie/api_except.py", line 1, in <module>
    from pyramid.httpexceptions import *
ImportError: No module named pyramid.httpexceptions
```

Zeitsperre:
If installing Magpie from Docker, the newest image available is based on "push-phoenix-btn", while the next supported image is "0.4.0" (23 March 2018). The current "latest" image is based on "Master" from around 6 months ago.

Should I trigger a new "latest" image based on Master?

Zeitsperre:
For reference: https://hub.docker.com/r/pavics/magpie/builds/

fmigneault:
"push-phoenix-btn" is under development for testing (not yet ready).
I would stick to 0.4.0 until this feature is complete

fmigneault:
resolved by PR #33

fmigneault:
resolved by PR #33

PR #33, opened 2018-05-08
add path to fix import

fmigneault:
resolves issue #27


dbyrns:
I would prefer if @huard could confirm that this PR solve his issue.

fmigneault:
@huard is the files permission not set locally for you? I don't see where the make install process would change permissions

huard:
Git stores the permissions, so my guess is that the Magpie devs worked on windows, which I believe has different default permissions from linux. 

I just think it's a matter of "chmod"ding the files and commit.

fmigneault:
That's why I'm a bit confused, on my side, it seems they are almost all set to -rwx------, except maybe a few files here and there that are -rwxrwxr-x. I can change all to -rw-rw-r--

huard:
image

fmigneault:
image

fmigneault:
I can at least chown the files for a common user, then try chmod them.

PR #35, opened 2018-05-11
Issue 30 login message failure

### LINKED ISSUE #30: Informative error message on login failure
At the moment we get 
```
{"reason": "not specified", "code": 401, "type": "application/json", "detail": "Login failure"}
```
I'd like to know whether it is because the user does not exist or the password is wrong, and eventually, an HTML page with a link to make a request to get an account or send a message to obtain the password.

fmigneault:
fixes #30

### DIFF
--- a/magpie/__meta__.py
+++ b/magpie/__meta__.py
@@ -3,7 +3,7 @@ General meta information on the magpie package.
 """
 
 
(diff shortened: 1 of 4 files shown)


fmigneault:
closes #30

Issue #37, opened 2018-05-14
Separate personal groups from standard groups


dbyrns:
magpie/ui/groups : 
- Should not show personal groups (solve also the problem that personal group should not be deleted directly (they are when deleting a user)

magpie/ui/users/add : 
- Combobox "user group" should not show personal groups

magpie/ui/users/username : 
- Groups Membership should not show personal groups (solve also the problem that a user should not be able to remove its own personal group)
- A panel showing personal group permission as in magpie/ui/groups/username/default should be shown (this is a more intuitive way to edit "user" permission. The fact that in the background we handle this by using a personnal group doesn't need to be shown in the UI)

fmigneault:
@dbyrns 
How should we handle following cases:

1. Creation of a new user, the "user group" dropdown list will have which default group, if any?
2. User "admin" is currently used as a joinable group my multiple other users, but is also an actual "user" on itself. Should a new group "administrators" be made for the purpose of joining/allowing admin rights and user "admin" remain a hidden special case?
3. User "anonymous" which is used as "not signed in", but is also a user internally... should it be hidden?

dbyrns:
1. I would say "users" (which means you can use the platform).
2. Good point, I like your solution but will need a thorough reflexion to avoid breaking something upon deployment.
3. No because we need it to allow permission editing.

So a fresh set of users/groups would be:
Users:
anonymous (with the hidden anonymous group)
admin (with the hidden admin group)

Groups:
users (standard permissions to use the platform), no members
administrators (permissions editing authorisation only), admin as member

As new users are created they will be either "users" or "administrators" or both. And some groups could be added as needed but I see a limited number of them without user "group" pollution.

fmigneault:
closed via PR #46

PR #46, opened 2018-05-30
separate user groups

fmigneault:
## API

- magpie v0.5.0
- no more 'personal' group for users
- auto migrate db to latest alembic version
- new routes for user permissions
    -  GET `/users/{user_name}/resources`
    -  GET `/users/{user_name}/inherited_resources`
    -  GET `/users/{user_name}/resources/{id}/permissions`
    -  GET `/users/{user_name}/resources/{id}/inherited_permissions`
    -  GET `/users/{user_name}/services`
    -  GET `/users/{user_name}/inherited_services`
    -  GET `/users/{user_name}/services/{svc}/permissions`
    -  GET `/users/{user_name}/services/{svc}/inherited_permissions`
- big cleanup of scripts, function locations and obsolete code

## UI
- checkbox for inherited user permissions or not
- editable permissions only in user-only view

addresses #37


magpie/magpie.py
```diff
@@ -65,37 +68,26 @@ def get_request_info(request, default_msg="undefined"):
...
-    from db import is_database_ready
-    if not is_database_ready():
+    # migrate db as required and check if database is ready
+    try:
+        db.run_database_migration()
+    except ImportError:
+        pass
+    except Exception as e:
+        raise Exception('Database migration failed [{}]'.format(str(e)))
+    if not db.is_database_ready():
```
dbyrns: The fct is_database_ready should not be called before doing the migration?

fmigneault: `is_database_ready` verifies that all db tables for Group, Users, Services, etc. have been created.
In the event that db migration requires a schema change, a missing table would always fail because migration didn't occur already.

dbyrns: Make sense. I was misinterpreting that function as a waiting point for the postgres instance to be ready which obviously is not.

fmigneault:
closes #39

Issue #40, opened 2018-05-22
`permission_names` not returned in most GET calls


fmigneault:
A few examples where `"permission_names": []`
- `{host_bird}/magpie/services/thredds`
- `{host_bird}/magpie/services/thredds/resources`
- `{host_bird}/magpie/services/types/wps`

Returned list in each case should be the list of available permissions (read, write, getcapabilities, and so on)

PR #49, opened 2018-06-05
Permissions refactoring for Service-typed Resources

fmigneault:
## Permissions changes:
- resources and services now display all **applicable**/**permitted** `permission_names` according to service-specific sub-resources permissions
- user and group resources/services still display only **applied** permissions (ie: same ones as ticked in UI)
- added more checks to restrict creation of unsupported resources under a given service
- added more checks to restrict assignation of unsupported permissions under a specific *service-typed* resource
- db migration for new `root_service_id` stored under every sub-resource
    * links are created for existing resources
    * created resources will be updated automatically with proper link to service
- magpie `0.5.1`

addresses issue #40 

## Others: 
- adjusted error message returned when user has insufficient permissions to execute a call, now 401 is returned (instead of 403) and is in JSON format.

### DIFF
--- a/magpie/__meta__.py
+++ b/magpie/__meta__.py
@@ -3,7 +3,7 @@ General meta information on the magpie package.
 """
 
 
-__version__ = '0.5.0'
+__version__ = '0.5.1'
 
 __author__ = "Francois-Xavier"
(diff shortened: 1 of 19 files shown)


fmigneault:
@dbyrns available on `colibri` for testing

PR #75, opened 2018-06-29
service 'access'

fmigneault:
## changes

- `ServiceAccess` with single permission `access` to control root path URL
   (sub routes under service are all evaluated only based on the root service user access)
- service new parameter `child_resource_allowed ` to block sub-resource creation under service


fmigneault:
@dbyrns je suis en train de documenter les issues trouvés reliés à ça pour garder un tracking

PR #102, opened 2018-09-13
add quote/bill stores for twitcher magpie adapter

fmigneault:
Adds default stores required by projects.


fmigneault:
will be added on twitcher-ogc side

PR #105, opened 2018-09-18
Fix assigning a user to a group in api and ui

davidcaron:
There was a bug in the UI and api when trying to add a user to a group in both the "edit group" view and the "edit user" view.

From the UI, the POST route was: `/users/{user_name}/groups/{group_name}`
I changed it to: `/users/{user_name}/groups` with the group_name in the post data

Also, I had to change the view of the `/users/{user_name}/groups` because it was expecting a group_name in the `matchdict`

### DIFF
--- a/magpie/api/management/user/user_views.py
+++ b/magpie/api/management/user/user_views.py
@@ -4,6 +4,7 @@ from magpie.api.api_requests import *
 from magpie.api.api_rest_schemas import *
 from magpie.api.management.user.user_formats import *
 from magpie.api.management.user.user_utils import *
+from magpie.api.management.group.group_utils import *
 from magpie.api.management.service.service_formats import format_service, format_service_resources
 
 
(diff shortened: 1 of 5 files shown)


davidcaron:
@fmigneault Please advise on adding a test for this...

magpie/api/management/user/user_views.py
```diff
@@ -99,7 +100,14 @@ def assign_user_group(request):
     """Assign a user to a group."""
     db = request.db
     user = get_user_matchdict_checked_or_logged(request)
-    group = get_group_matchdict_checked(request)
+
+    group_name = get_value_multiformat_post_checked(request, 'group_name')
+    group = evaluate_call(lambda: zig.GroupService.by_group_name(group_name, db_session=request.db),
```
fmigneault: I don't see where this `zig` namespace is defined. Where is it coming from?

davidcaron: It's actually defined in api_requests.py... `zig` is used in other places in `user_views.py` so I used the same reference here. I could remove the `zig.` part and it works just fine. I think we could talk about cleaning these imports in another issue/PR, but for now I just used the same reference as other places in the same file.

PR #106, opened 2018-09-25
Add label to display resources

### LINKED ISSUE #98: Add label to resources in magpie
For some services, resources could be numerical with no context (project-api is a good example). 
For them we could assign to the resource a friendly name that would be shown in the magpie UI.

Since the friendly name could be changed on the remote server it should only be store on the remote resource table which is update periodically and use for display only. Thus I see no needs to update the API or the main resources and permissions tables.

davidcaron:
Depends on PR #104.
Fixes #98.

@fmigneault, I would need some guidance in swagger to add the optional `resources_display_name` argument to the POST request at path `/resources`.


davidcaron:
@fmigneault I think I figured it out in c1685e1 how to modify swagger.

This is all functional now, as far as I know.

fmigneault:
will wait on PR#104 to be merged, then rebase on top, so that there's less lines to go back through

PR #109, opened 2018-10-02
Service API ponctual permissions

fmigneault:
depends on PR #104 

* adds the `read-match` and `write-match` permissions to `ServiceAPI` route resources.
  requested route must be an exact match to ticked permission to pass
* `read`/`write` permissions continue to englobe all sub-resources 
  ie: `read` on `/processes` will grant `read` on `/processes/...`
* use `/users/{id}/services?cascade=True` to return services if any of its sub-resources has user permissions
    - otherwise default behaviour remains (direct permissions on service must be set for user to return a given service)
    - can be combined with `inherit` query

* start to deprecate `/users/.../inherited_<>` routes (still available, but will log warnings and such)
* use `/users/.../<>?inherit=True` instead to return `<>` inheriting user's groups permissions 

* change to magpie `0.7.0` since including big change of sync remote

## To Do

tag `0.7.0` when this PR, PR #104, PR #106 are merged


fmigneault:
Tomorrow, we will need to give a try to a massive merge of all these PRs.

dbyrns:
It will be epic !

davidcaron:
I will try go ahead with the merge in a new branch and deploy on colibri if everyone is ok

dbyrns:
Good!

tomLandry:
Will you tell me about these tales of merging epicness?

fmigneault:
closed via #110

Issue #139, opened 2019-01-30
Continuous Integration


fmigneault:
- Travis-CI (https://travis-ci.com/Ouranosinc/Magpie)
- Implement CI pipeline with releases to docker hub (https://hub.docker.com/r/pavics/magpie)
- Run unittests on PR, tags, etc.  (ref #74, #120, #130)

fmigneault:
tests are executed on travis-ci
dockerhub does the auto-fetch of tags and builds them

PR #140, opened 2019-01-30
version 0.9.0 improvements

fmigneault:
* add permissions config to auto-generate user/group rules on startup
* fix many invalid or erroneous swagger specifications
* attempt db creation on first migration if not existing
* add continuous integration testing and deployment (#139)
* reduce excessive sqlalchemy logging using `MAGPIE_LOG_LEVEL >= INFO`


Makefile
```diff
@@ -64,91 +101,183 @@ clean-test:
...
 	@echo "Starting Cron service..."
 	cron
 
 .PHONY: start
-start: cron install
+start: install
 	@echo "Starting Magpie..."
-	exec gunicorn -b 0.0.0.0:2001 --paste "$(CUR_DIR)/magpie/magpie.ini" --workers 10 --preload
+	@bash -c 'source "$(CONDA_HOME)/bin/activate" "$(CONDA_ENV)"; \
+		exec gunicorn -b 0.0.0.0:2001 --paste "$(CUR_DIR)/magpie/magpie.ini" --workers 10 --preload &'
```
dbyrns: That little & at the end of the command looks like a hack, no?

magpie/register.py
```diff
@@ -25,6 +49,11 @@
 SERVICES_PHOENIX = 'PHOENIX'
```
dbyrns: Is this file only related to Phoenix? Because all Phoenix related stuff could be marked as deprecated... and removed as soon as the latest twitcher version is used.

Makefile
fmigneault: Tells bash to not wait for the return of the call.
So basically, run the process detached.

magpie/register.py
fmigneault: Indeed, all of Phoenix related code is preserved for back compatibility to old twitcher, but is mostly deprecated now.

PR #143, opened 2019-02-11
Docker and others

davidcaron:
Docker:

- image based on alpine
- smaller image (150 Mb vs 800 Mb)
- faster incremental builds (don't reinstall dependencies)
- faster boot time (don't check dependencies at startup)

When the database doesn't exist, create it using sqlalchemy_utils. Relying on postgresql user postgres being present is not always reliable. This library can create the database without it.

Other small fixes (see commit messages)


fmigneault:
closed via #150

PR #146, opened 2019-02-15
Gunicorn access logs

davidcaron:
This prints the gunicorn access logs to stdout.

I don't know what you think, but I think this helps debugging...

### DIFF
--- a/magpie/magpie.ini
+++ b/magpie/magpie.ini
@@ -52,6 +52,7 @@ document_root = %(here)s/ui/swagger
 [server:main]
 use = egg:gunicorn#main
 host = localhost
+logconfig = %(here)s/magpie.ini
 port=2001
 timeout=10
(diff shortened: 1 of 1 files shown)


fmigneault:
@davidcaron 
having quite a lot of trouble figuring out what's failing on travis
everything works fine for local/remote tests on my machine

tomLandry:
Maybe the type of thing @ldperron could help troubleshoot?

ldperron:
Unfortunately, Travis doesn't collect any details about the 500 error. Should we try adding it to Sentry?

davidcaron:
I restarted it and it worked... 

The only thing I see is that when I created the PR, it targeted the master branch, but I quickly changed it to login-local-subreq. Maybe travis picked up something wrong when checking out the PR.

PR #152, opened 2019-02-22
Docker build

davidcaron:
CHANGE:
- in the Dockerfile, don't reinstall dependencies when there is only a code change

FIX:
- futures is required for gunicorn threads in python 2.7
- add configparser for python 2.7
- change MAGPIE_DIR from ARG to ENV in Dockerfile
- create_user_resource_permission takes a models.User, not a user id

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,8 @@
 FROM python:2.7-alpine
 MAINTAINER Francis Charette-Migneault
 
-ARG MAGPIE_DIR=/opt/local/src/magpie
+# the cron service depends on the $MAGPIE_DIR environment variable
+ENV MAGPIE_DIR=/opt/local/src/magpie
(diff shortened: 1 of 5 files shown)


fmigneault:
@davidcaron 
configparser is added via six which handles the py2/3 SafeConfigParser/ConfigParser variantions.
where is the requirement problem? maybe just need to adjust the import to use six?
can merge after this adjustment

davidcaron:
> @davidcaron
> configparser is added via six which handles the py2/3 SafeConfigParser/ConfigParser variantions.
> where is the requirement problem? maybe just need to adjust the import to use six?
> can merge after this adjustment

@fmigneault I removed the requirement and imported configparesr from six.moves. There was an `import configparser` statement in `magpie/common.py`

PR #155, opened 2019-03-01
magpie static & url prefix fixes

### LINKED ISSUE #153: Static files are not served properly behind nginx
When magpie is behind an nginx reverse proxy served at an url like: 
https://192.168.99.201/magpie/
In the ui, an example request for an asset looks like:
https://192.168.99.201/static/magpie.png

So it should be:
https://192.168.99.201/magpie/static/magpie.png

I think this should be modified in magpie, and not in nginx, because redirecting /static to /magpie/static could be in conflict with another service relying on the /static route.

fmigneault:
resolves #153


fmigneault:
tested and deployed on https://geoimagenetdev.crim.ca/magpie (0.9.4-dev / pseudo 0.9.5)
closes #155

PR #191, opened 2019-06-21
Add return statement

dbyrns:
Found that while setupping a dev env. Probably never caused an issue before because the MAGPIE_LOG_LEVEL was always defined.


fmigneault (approved):
@dbyrns 
I've added the changes returning `Permission` enum as discussed.

PR #203, opened 2019-07-05
set `use_tweens` to True when using the `invoke_subrequest` pyramid method

davidcaron:
I could reproduce the connection pool issues we were having, and this should fix it. I think multiple threads were getting the same session from the factory, and the sessions were then probably not closed properly.

Also, remove pool_threadlocal which is equivalent to use_threadlocal, which is deprecated.

Explanation here: https://docs.sqlalchemy.org/en/13/orm/contextual.html#thread-local-scope

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -6,6 +6,12 @@ History
 Unreleased
 ---------------------
 
+Features / Changes
+~~~~~~~~~~~~~~~~~~~~~
+* add ``use_tweens=True`` to ``request.invoke_subrequest`` calls in order to properly handle the nested database
+  transaction states with the manager (#203). Automatically provides ``pool_threadlocal`` functionality added in
+  ``1.3.1`` as per implementation of ``pyramid_tm`` (#201).
+
 1.3.1 (2019-07-05)
 ---------------------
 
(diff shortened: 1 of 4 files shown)


davidcaron:
@dbyrns, @fmigneault I think I found it... when magpie used the `invoke_subrequest` method of `pyramid`, the session used within this subrequest was not properly managed by `pyramid_tm` because `invoke_subrequest` doesn't use tweens by default, and the connection was left in a transaction state.

I think using `pool_threadlocal` fixed it because the session obtained by the subrequest was the same as in the main request.

davidcaron:
Oh, and by default `pyramid_tm` uses thread locals, so as long as we let it handle the request, using multiple threads should be ok.

dbyrns:
So, setting pool_threadlocal to true isn't only pushing the limit ahead, it really removes the limit and we can  do as many subrequest as required without exhausting the connections, right?

davidcaron:
Yes, by setting `pool_threadlocal` to True, subrequests made with with `invoke_subrequest` will use the same database connection from the pool. I think this translates to a nested transaction. The main request has a transaction, and the subrequests each have a nested transaction within the main one.

PR #205, opened 2019-08-08
fix migraiton script

fmigneault:
Fixes the error encountered during migration.
Using the db you provided, I reach the "latest" version and Magpie is booted.

Fix for Ouranosinc/PAVICS#149


fmigneault:
tag `1.3.4` is pushed, should be available after some time in dockerhub

PR #218, opened 2019-09-11
Use pyramid-beaker to cache acl permissions for a specific service and user name

davidcaron:
I only tested this with the `ServiceAPI` service type. 

Basically, the permissions are cached for a specific service and user. When there are a lot of concurrent requests, adding a cache with a small expiration time of 5 seconds can make a big difference (2-3 times more requests handled per second).

@fmigneault I don't think so, but look if I didn't miss anything regarding permissions that caching could interfere with...

### DIFF
--- a/magpie/adapter/__init__.py
+++ b/magpie/adapter/__init__.py
@@ -150,6 +150,7 @@ class MagpieAdapter(AdapterInterface, Singleton):
 
         LOGGER.info("Loading MagpieAdapter config")
         config = get_auth_config(container)
+        config.include("pyramid_beaker")
 
         # use pyramid_tm to hook the transaction lifecycle to the request
         # make request.db available for use in Pyramid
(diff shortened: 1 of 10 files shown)


magpie/adapter/__init__.py
```diff
@@ -138,6 +139,7 @@ def owsproxy_config(self, container):
 
     def configurator_factory(self, container):
         settings = get_settings(container)
+        set_cache_regions_from_settings(settings)
```
fmigneault: Probably we should also add this call inside `magpie.app.main` since `@cache_region` affects generic services.

davidcaron:
> Also, we need to add the config in Twitcher's ini to make them work in the gateway.

Does twitcher needs to know about beaker? I wouldn't think so, as it's not a required dependency.

> I would also add a _warning_ note about caching timeout as it can make unexpected behaviours quite easily if set too high.

I set it up so that no caching is done by default... where should the warning be? In the configuration file or in the docs?

PR #219, opened 2019-09-16
Create users script

ChaamC:
Script to create or delete a list of users


processUsers.py
```diff
@@ -0,0 +1,90 @@
...
+    users = []
+    for user in user_names:
+        response = session.delete(magpie_url + "/users/" + user)
+        users.append({'user_name': user, 'result': format_response(response)})
+    return users
+
+
+if __name__ == '__main__':
+    email_list = sys.argv[1:]
+    magpie_url = "http://192.168.99.201:2001/magpie"
```
dbyrns: Could use localhost rather than a specific ip here

processUsers.py
```diff
@@ -0,0 +1,90 @@
...
+        response = session.delete(magpie_url + "/users/" + user)
+        users.append({'user_name': user, 'result': format_response(response)})
+    return users
+
+
+if __name__ == '__main__':
+    email_list = sys.argv[1:]
+    magpie_url = "http://localhost/magpie"
+    magpie_admin_user_name = "admin"
+    magpie_admin_password = "admin"
```
fmigneault: These parameters should all be passed as input to the script.
Check `argparse` module to make a clean CLI.

processUsers.py
```diff
@@ -0,0 +1,90 @@
...
+        output = "USERNAME".ljust(COLUMN_SIZE) + "PASSWORD".ljust(COLUMN_SIZE) + "RESULT".ljust(COLUMN_SIZE) + "\n"
+        output += "".ljust(COLUMN_SIZE * 3, "_") + "\n\n"
+        for user in users:
+            output += user['user_name'].ljust(COLUMN_SIZE) + user['password'].ljust(COLUMN_SIZE) + user['result'].ljust(
+                COLUMN_SIZE) + "\n"
+
+        filename = "createUsers_log__" + datetime.datetime.now().strftime('%Y%m%d__%H%M%S') + ".txt"
+        file = open(filename, "w+")
+        file.write(output)
+        file.close()
```
fmigneault: use a `with(...)` block to auto handle closing file in case of error

processUsers.py
```diff
@@ -0,0 +1,90 @@
...
+
+        print(output)
+        print("Output results sent to " + filename)
```
fmigneault: Replace all `print` by a logger with `logging` module. (see `magpie.utils` for examples of simple setup of a logger).
Makes it easier to update where/how output info is provided when deployed somewhere using a logging config.

processUsers.py
```diff
@@ -0,0 +1,90 @@
+#!/usr/bin/env python3
```
fmigneault: rename the file `create_users.py` and move it under `magpie/helpers`

PR #221, opened 2019-10-01
Fix migration: don't use the orm

davidcaron:
We can't query using models.Resource.resource_type because it's a declared_attr in sqlalchemy. 

In general, migrations that don't depend on the sqlalchemy models will be more stable as the models get modified.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -6,6 +6,10 @@ History
 Unreleased
 ---------------------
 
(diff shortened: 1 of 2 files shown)


davidcaron:
> Was it was tested on your side with migration going through this version and actually updating an existing project-api resource?

yes, and the `if` case was also hit.

PR #222, opened 2019-10-02
The request parser should default to 'application/json' content type only when the request has a body

davidcaron:


magpie/owsrequest.py
```diff
@@ -14,7 +14,7 @@
 import lxml.etree
 import json
 if TYPE_CHECKING:
-    from requests import Request  # noqa: F401
+    from pyramid.request import Request
```
fmigneault: not sure if this is the cause of failing tests...

magpie/owsrequest.py
```diff
@@ -26,31 +26,29 @@ def ows_parser_factory(request):
     Default to JSON if no ``Content-Type`` is specified or if it is 'text/plain' but can be parsed as JSON. Otherwise,
     use the GET/POST WPS parsers.
     """
-    content_type = get_header("Content-Type", request.headers, default=None, split=";,")
-    if content_type is None or content_type == CONTENT_TYPE_PLAIN:
+    content_type = request.headers.get("Content-Type", None)
```
fmigneault: use `get_header`
ensure compat where requests or pyramid.request is used

magpie/owsrequest.py
```diff
@@ -26,31 +26,29 @@ def ows_parser_factory(request):
     Default to JSON if no ``Content-Type`` is specified or if it is 'text/plain' but can be parsed as JSON. Otherwise,
     use the GET/POST WPS parsers.
     """
```
fmigneault: adjust docstring to reflect change (ie: attempt parsing JSON if body is also available)

magpie/owsrequest.py
```diff
@@ -26,31 +26,29 @@ def ows_parser_factory(request):
...
-            if request.body:
-                # raises if parsing fails
-                # noinspection PyUnresolvedReferences
-                json.loads(request.body)
-                content_type = CONTENT_TYPE_JSON
+            json.loads(body)
         except ValueError:
-            pass
+            return False
+        return True
```
fmigneault: move to `magpie.utils`

magpie/owsrequest.py
```diff
@@ -26,31 +26,29 @@ def ows_parser_factory(request):
...
+
+    if content_type is None or content_type == CONTENT_TYPE_PLAIN:
+        if is_json_body(request.body):
+            # default to json when can be parsed as json
+            request.content_type = content_type = CONTENT_TYPE_JSON
```
fmigneault: not sure if this is requests / pyramid.request compatible...
do both have `.content_type` property?
maybe use `setattr` (to test ?)

davidcaron: In `magpie.api.requests.get_multiformat_any`, the `request.content_type` property is accessed. So it's already not `requests` compatible. In which case does this function receive a `requests.Request` object? I would believe it's always a pyramid request.

fmigneault: I'm not 100% sure but I remember having multiple places where both where received according to test cases and Twitcher integration. This is why there are multiple util functions to handle both.

davidcaron: how about `request.headers["Content-Type"] = request.content_type = content_type = CONTENT_TYPE_JSON`?

fmigneault: Probably. 
I have no problem with it if it works as is and all tests pass.

PR #223, opened 2019-10-11
Fix #165 (database version)

### LINKED ISSUE #165: db migration alembic version failing on travis-ci
Alembic db version fetch is discarded on raised error since https://github.com/Ouranosinc/Magpie/commit/6435886da103b0ca60870eb29707cd17f8d11893 to allow tests to pass because it fails to be retrieved on travis-ci.

Test Failing: https://travis-ci.com/Ouranosinc/Magpie/builds/106243123
Test Success: https://travis-ci.com/Ouranosinc/Magpie/builds/106247111
Reference: #156

Even when changing `magpie.db_migration=True` during tests, table `alembic_version` seems to be missing or queries to it fail. 
Migration doesn't seem to fail when enabled, as it does not report any issue/warning/error at first glance. 
Fetching of `alembic_version` works on local instance and deployed instance.

davidcaron:
Fix for warning: "Textual SQL expression 'alembic_version' should be explicitly
declared as text('alembic_version')"

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -6,6 +6,10 @@ History
 Unreleased
 ---------------------
 
+Bug Fixes
+~~~~~~~~~~~~~~~~~~~~~
+* fix the alembic database version number in the /version route
+
 1.6.2 (2019-10-04)
 ---------------------
 
--- a/magpie/db.py
+++ b/magpie/db.py
@@ -114,7 +114,7 @@ def run_database_migration(db_session=None):
 
 def get_database_revision(db_session):
     # type: (Session) -> Str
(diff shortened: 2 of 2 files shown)


fmigneault:
@davidcaron 
Does it only fix the warning or #165 is also fixed?
By this, I mean the `version=null` that is used by default in `GET /magpie/version` response because lookup of `alembic_version` fails.

davidcaron:
Yes, the /version route returns the proper alembic version number.

fmigneault:
@davidcaron Nice.
Can you add a fix #165 note in the changelog?
After good to merge.

PR #228, opened 2019-11-05
Fix resource sync

### LINKED ISSUE #226: cronjob to sync resources needs to be updated to use alpine cron in docker image
Logging and file paths need to be modified, the defaults for cron on alpine are:

``` bash
bash-5.0# crond --help 
BusyBox v1.30.1 (2019-06-12 17:51:55 UTC) multi-call binary. 

Usage: crond -fbS -l N -d N -L LOGFILE -c DIR

        -f      Foreground
        -b      Background (default)
        -S      Log to syslog (default)
        -l N    Set log level. Most verbose 0, default 8
        -d N    Set log level, log to stderr
        -L FILE Log to FILE
        -c DIR  Cron dir. Default:/var/spool/cron/crontabs
```

davidcaron:
Fix #226

I found a couple bugs along the way, see the commits messages.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,8 +17,6 @@ RUN chmod 0644 $CRON_DIR/magpie-cron
 COPY magpie/__init__.py magpie/__meta__.py $MAGPIE_DIR/magpie/
 COPY requirements* setup.py README.rst HISTORY.rst $MAGPIE_DIR/
 
-# Note: the requests package needs to be installed before threddsclient
-# this is a bug that should be fixed in threddsclient>0.3.5
 RUN apk update \
     && apk add \
     bash \
@@ -32,7 +30,7 @@ RUN apk update \
     python-dev \
     musl-dev \
     postgresql-dev \
-    && pip install --no-cache-dir --upgrade pip setuptools requests \
+    && pip install --no-cache-dir --upgrade pip setuptools \
     && pip install --no-cache-dir -e $MAGPIE_DIR \
     && apk --purge del .build-deps
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,6 @@ ENV CRON_DIR=/etc/crontabs
 # magpie cron service
 COPY magpie-cron $CRON_DIR/magpie-cron
 RUN chmod 0644 $CRON_DIR/magpie-cron
-RUN touch /var/log/cron.log
 
 COPY magpie/__init__.py magpie/__meta__.py $MAGPIE_DIR/magpie/
 COPY requirements* setup.py README.rst HISTORY.rst $MAGPIE_DIR/
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,17 +22,17 @@ COPY requirements* setup.py README.rst HISTORY.rst $MAGPIE_DIR/
 # this is a bug that should be fixed in threddsclient>0.3.5
 RUN apk update \
     && apk add \
-        bash \
-        postgresql-libs \
-        py-pip \
-        libxslt-dev \
+    bash \
+    postgresql-libs \
+    py-pip \
+    libxslt-dev \
     && apk add --virtual .build-deps \
-        supervisor \
-        gcc \
-        libffi-dev \
-        python-dev \
-        musl-dev \
-        postgresql-dev \
+    supervisor \
+    gcc \
+    libffi-dev \
+    python-dev \
+    musl-dev \
+    postgresql-dev \
     && pip install --no-cache-dir --upgrade pip setuptools requests \
     && pip install --no-cache-dir -e $MAGPIE_DIR \
     && apk --purge del .build-deps
@@ -42,4 +42,4 @@ COPY ./ $MAGPIE_DIR
 # equivalent of `make install` without conda env and pre-installed packages
 RUN pip install --no-dependencies -e $MAGPIE_DIR
 # equivalent of `make cron start` without conda env
-CMD crond -c $CRON_DIR && gunicorn -b 0.0.0.0:2001 --paste $MAGPIE_DIR/config/magpie.ini --workers 10 --preload
+CMD crond -c $CRON_DIR && gunicorn -b 0.0.0.0:2001 --paste $MAGPIE_CONFIG_DIR/config/magpie.ini --workers 10 --preload
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,16 +8,18 @@ ENV MAGPIE_DIR=/opt/local/src/magpie
 ENV MAGPIE_ENV_DIR=$MAGPIE_DIR/env
 WORKDIR $MAGPIE_DIR
 
+ENV CRON_DIR=/etc/crontabs
+
 # magpie cron service
-COPY magpie-cron /etc/cron.d/magpie-cron
-RUN chmod 0644 /etc/cron.d/magpie-cron
-RUN touch ~/magpie_cron_status.log
-# set /etc/environment so that cron runs using the environment variables set by docker
-RUN env >> /etc/environment
+COPY magpie-cron $CRON_DIR/magpie-cron
+RUN chmod 0644 $CRON_DIR/magpie-cron
+RUN touch /var/log/cron.log
 
 COPY magpie/__init__.py magpie/__meta__.py $MAGPIE_DIR/magpie/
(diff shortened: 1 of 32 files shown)


Dockerfile
```diff
@@ -8,16 +8,18 @@ ENV MAGPIE_DIR=/opt/local/src/magpie
...
+
 # magpie cron service
-COPY magpie-cron /etc/cron.d/magpie-cron
-RUN chmod 0644 /etc/cron.d/magpie-cron
-RUN touch ~/magpie_cron_status.log
-# set /etc/environment so that cron runs using the environment variables set by docker
-RUN env >> /etc/environment
+COPY magpie-cron $CRON_DIR/magpie-cron
+RUN chmod 0644 $CRON_DIR/magpie-cron
+RUN touch /var/log/cron.log
```
fmigneault: Will not match existing variable `MAGPIE_CRON_LOG`.

magpie-cron
```diff
@@ -1 +1 @@
-0 * * * * root /bin/bash -c "set -a ; . $MAGPIE_DIR/env/magpie.env ; . $MAGPIE_DIR/magpie/env/magpie.env ; . $MAGPIE_DIR/env/postgres.env ; . $MAGPIE_DIR/magpie/env/postgres.env ; set +a ; python -c 'from magpie.helpers.sync_resources import main; main()'" > ~/magpie_cron_status.log 2>&1
+0 * * * * /bin/bash -c "set -a ; source <($MAGPIE_DIR/env/*.env) ; set +a ; python -c 'from magpie.helpers.sync_resources import main; main()'" >> /var/log/cron.log 2>&1
```
fmigneault: Should use `MAGPIE_ENV_DIR` in case it is overridden.

Dockerfile
```diff
@@ -40,4 +42,4 @@ COPY ./ $MAGPIE_DIR
 # equivalent of `make install` without conda env and pre-installed packages
 RUN pip install --no-dependencies -e $MAGPIE_DIR
 # equivalent of `make cron start` without conda env
-CMD crond && gunicorn -b 0.0.0.0:2001 --paste $MAGPIE_DIR/config/magpie.ini --workers 10 --preload
+CMD crond -c $CRON_DIR && gunicorn -b 0.0.0.0:2001 --paste $MAGPIE_DIR/config/magpie.ini --workers 10 --preload
```
fmigneault: Should use `MAGPIE_CONFIG_DIR` in case it is overridden.

magpie/api/management/resource/resource_utils.py
```diff
@@ -115,7 +115,7 @@ def get_resource_permissions(resource, db_session):
     ax.verify_param(resource, notNone=True, httpError=HTTPBadRequest, paramName=u"resource",
                     msgOnFail=s.UserResourcePermissions_GET_BadRequestResourceResponseSchema.description)
     # directly access the service resource
-    if resource.root_service_id is None:
+    if resource.root_service_id is None or resource.root_service_id == resource.resource_id:
```
fmigneault: Was this case encountered somewhere?
`resource.root_service_id` is expected to always be `None` if it is a service.
If this is not respected, it's an implementation error somewhere else that must be fixed as it could break other code logic.

magpie/api/management/resource/resource_views.py
```diff
@@ -59,6 +59,10 @@ def create_resource_view(request):
     resource_display_name = ar.get_multiformat_any(request, "resource_display_name", default=resource_name)
     resource_type = ar.get_value_multiformat_post_checked(request, "resource_type")
     parent_id = ar.get_value_multiformat_post_checked(request, "parent_id")
+    try:
+        parent_id = int(parent_id)
+    except ValueError:
+        pass
```
fmigneault: This feels off. If the id is invalid, it should probably fail directly instead of silently passing.
Invalid id will simply be forwarded to `create_resource`.
Is this the possible source of the previous comment about `resource_id != None` for a service?

magpie/ui/management/views.py
```diff
@@ -19,16 +19,17 @@
...
 import json
 import six
 if TYPE_CHECKING:
     from magpie.definitions.sqlalchemy_definitions import Session  # noqa: F401
     from magpie.definitions.typedefs import List, Optional  # noqa: F401
-
 LOGGER = get_logger(__name__)
 
 
+
```
fmigneault: remove extra space

requirements.txt
```diff
@@ -24,7 +24,7 @@ simplejson
 six
 sqlalchemy
 sqlalchemy_utils
-threddsclient==0.3.4
+threddsclient==0.3.5
```
fmigneault: Add `FIXME` note corresponding to Dockerfile `request` installation order when updated with fix mentioned.

magpie/api/management/resource/resource_views.py
davidcaron: I agree it feels off. But can you suggest another way to do it?
The problem is that the parent_id is a string according to the schema. Magpie calls this route itself in a couple places, and sometimes sends a string, sometimes an integer. `create_resource` in `magpie/api/management/resource/resource_utils.py` expects an integer, and returns the proper error message when the parent_id is invalid.

magpie/api/management/resource/resource_utils.py
davidcaron: Hum... I think I might have to do a small database migration here. In my last PR, I fixed a database migration, but I thought that setting the `root_service_id` to the `resource_id` for the services was the intention. (see `a395ef9d3fe6_reference_root_service.py`)

I could add the migration in this PR and remove this check.

Dockerfile
davidcaron: Look at the `CRON_DIR` variable in the Dockerfile, I think `MAGPIE_CONFIG_DIR` refers to something else.

Dockerfile
davidcaron: `/var/log/cron.log` is known at build time, while `MAGPIE_CRON_LOG` can be changed at runtime... I think these don't have to match? 
Maybe we just need one log file... I think MAGPIE_CRON_LOG wouldn't be necessary, what do you think?

davidcaron:
Note regarding the problem when pip installing `threddsclient`: the order of the dependencies when running `pip install` seems to be non deterministic. When running `pip install magpie` in a fresh environment, it works, but when building the docker image, we need to install `requests` first for it to work.

magpie/api/management/resource/resource_views.py
fmigneault: Then we should update the schema to require an int.
Where is this route called that a string is passed? These should be fixed instead.

magpie/api/management/resource/resource_utils.py
fmigneault: Ok, yes please adjust within this PR.

Dockerfile
fmigneault: I meant in the `--paste` parameter.

Dockerfile
fmigneault: I don't mind removing it, but if it remains, it should be respected here and inside `magpie-cron` file.
Currently the variable is used here :
https://github.com/Ouranosinc/Magpie/blob/e8a47b0227910703201b4a159ad182c0270bd032/magpie/helpers/sync_resources.py#L336
So we would have different logging behaviours depending on which process runs the cron operation.

davidcaron: I ended up keeping only `MAGPIE_CRON_LOG`

magpie/api/management/resource/resource_views.py
davidcaron: Take a look at https://github.com/Ouranosinc/Magpie/pull/228/commits/44bc571e10b0954b9969f4b772d8773ee854a63c

magpie/api/management/resource/resource_utils.py
davidcaron: done in https://github.com/Ouranosinc/Magpie/pull/228/commits/f6d7700f567bed88201db57c9f0a21e4913b3b9a

davidcaron:
It seems good to go on my side, I`ll let you take a look at the modifications.

fmigneault:
Still has a .dockerignore conflict.
After its fixed, let travis test everything. 
Then good to merge if all tests passed.

fmigneault:
@davidcaron 
Add this step before install in makefile to fix the version conflict
https://github.com/pypa/pip/issues/4537#issuecomment-373386803

PR #232, opened 2019-11-11
fix magpie load config from dir

fmigneault:


tlvu:
For the uninitiated, can you explain what was the "restrictive file check"?

magpie/register.py
```diff
@@ -503,12 +503,18 @@ def _get_all_configs(path_or_dict, section):
...
+        list of configurations loaded if input was a directory path
+        list of single configuration if input was a file path
+        list of single configuration if input was a JSON dict
+        empty list if none of the other cases where matched
     """
     if isinstance(path_or_dict, six.string_types) and os.path.isdir(path_or_dict):
         dir_path = os.path.abspath(path_or_dict)
         return [_load_config(os.path.join(dir_path, f), section) for f in os.listdir(dir_path) if f.endswith(".cfg")]
-    return [_load_config(path_or_dict, section)]
+    elif isinstance(path_or_dict, dict):
```
davidcaron: I think you're missing another if case: `elif os.path.isfile(path_or_dict)`
If this function receives a path to a file, it will not be loaded.

fmigneault: Indeed. Nice catch.

fmigneault:
> For the uninitiated, can you explain what was the "restrictive file check"?

@tlvu 
Sorry for slow answer, I didn't realize this comment before.
Basically, the app calling the config loader that now accepts dir/file paths was already doing an `isfile` check, so dir paths where never actually trickled down to the function to load them.

tests/test_register.py
```diff
@@ -318,3 +321,32 @@ def test_variable_expansion_permissions_config_style(self):
...
         assert config["permissions"][1]["permission"] == "getcapabilities"
         assert config["permissions"][1]["group"] == admins
+
+    def test_get_all_config_from_dir(self):
+        with tempfile.TemporaryDirectory() as tmp_dir:
+            with tempfile.NamedTemporaryFile(mode="w", suffix=".cfg", dir=tmp_dir) as tmp1, \
+                 tempfile.NamedTemporaryFile(mode="w", suffix=".cfg", dir=tmp_dir) as tmp2:
+                # format doesn't matter
+                tmp1.write(json.dumps({"permissions": [{"perm": "permission1"}, {"perm": "permission2"}]}))
+                tmp1.seek(0)  # back to start since file still open (auto-delete if closed)
```
davidcaron: maybe what you want is `tmp1.flush()`?

fmigneault: maybe does the same, but using `seek(0)` as defined in tempfile doc

PR #238, opened 2019-11-19
Fix 500 on current user services

davidcaron:


davidcaron:
fixes https://github.com/Ouranosinc/PAVICS-DataCatalog/issues/50

huard:
@tlvu If the catalog_search notebook runs with this, we're good.

davidcaron:
> 
> 
> @tlvu If the catalog_search notebook runs with this, we're good.

Good to know, I'll test this one.

tlvu:
In case you want to do a real Jenkins test:
* deploy a new Magpie containing this fix
* request a build on Jenkins pointing to the server with this Magpie fix

PR #240, opened 2019-12-02
sort cfg by name for ordered loading

fmigneault:


magpie/register.py
```diff
@@ -510,13 +510,13 @@ def _get_all_configs(path_or_dict, section):
...
     .. note::
-        Order of file loading is not guaranteed if specifying a directory path.
+        Order of file loading will be resolved by alphabetically sorted filename if specifying a directory path.
     """
     if isinstance(path_or_dict, six.string_types):
         if os.path.isdir(path_or_dict):
             dir_path = os.path.abspath(path_or_dict)
-            return [_load_config(os.path.join(dir_path, f), section)
-                    for f in os.listdir(dir_path) if f.endswith(".cfg")]
+            cfg_names = list(sorted({fn for fn in os.listdir(dir_path) if fn.endswith(".cfg")}))
```
tlvu: I think sorted already return a list so you do not have to make a list again.

fmigneault: Returns an iterator.
Though since it is iterated over right after, list conversion is indeed not required here.

PR #259, opened 2020-03-09
accept header checks

fmigneault:
allow more accept header content-type if ui/static route is requested for app mounted behind proxy with additional prefix than just /magpie


fmigneault:
@ldperron just as additional measure of the problem encountered, maybe it can help avoid it without further hacking around of the server config

PR #340, opened 2020-07-24
Milestone 2.0 - User scope views, Discoverable Groups & Patches

### LINKED ISSUE #337: [BUG] mismatch between documentation and requests
Documentation indicates some 'current' user routes as valid (example below, maybe others also)
https://pavics-magpie.readthedocs.io/en/latest/api.html#tag/User%2Fpaths%2F~1users~1%7Buser_name%7D~1groups%2Fpost

But this is clearly forbidden here: 
https://github.com/Ouranosinc/Magpie/blob/10849ba702772fb336b1b60c9f9010b421b4d6d4/magpie/api/management/user/user_views.py#L67-L70

fmigneault:
## summary of main features and changes

- add discoverable groups (publicly available for self-registration by user to become member of that group)
- add user-scoped views that allow operations such as self-edit user information & self-registration to discoverable groups
- add +120 test cases for new features & others for existing features that were not fully tested
- improve UI with many rendering tweaks (eg: modified theme, display version information, etc.)
- display errors with UI page when API results into invalid operation (e.g.: unauthorized)
- multiple fixes to expected behavior but previously not completely fulfilled as intended (validated by new tests)
- reduce produced outputs for non-admin users to avoid "leaking" some details they don't need (should not) know
- obsolete many routes formats that where deprecated for an extremely long time
- convert `PUT` to `PATCH` since this is the RESTful method for partial update (relates to #349)
- many docs update for permissions and explanations of features

##  full changelog (quite massive, see for details)

https://github.com/Ouranosinc/Magpie/blob/user-scope-views/CHANGES.rst#200-2020-07-31

### other 
- close #337, user will be able to update its own email & password, username still not permitted as it acts as 'id' in request paths


magpie/alembic/versions/03b54feffe45_ensure_anonymous_group_applied_to_users.py
```diff
@@ -35,17 +60,23 @@ def upgrade():
...
+            session.execute(groups.insert().values(group_name=anonym_name, member_count=len(all_user_ids)))
+            anonym_group = session.execute(query).fetchone()
+
+        query = sa.select([users_groups.c.user_id]).where(users_groups.c.group_id == anonym_group.id)
+        users_with_anonym = session.execute(query)
+        users_with_anonym = set(users_with_anonym or [])  # handle if None
+        missing_grp_users = all_user_ids - users_with_anonym
+        missing_grp_users = [{"user_id": usr_id, "group_id": anonym_group.id} for usr_id in missing_grp_users]
+        if missing_grp_users:
+            session.execute(users_groups.insert(), missing_grp_users)
```
dbyrns, line 79: This comment apply to all alembic patch update. Apart from the discoverable group patch which is new and will be applied, how can we know that the old patches are still working after modifications? Do they are all executed for a new deployment? Are they part of a test procedure?

dbyrns (commented):
Review in progress (33%)

magpie/ui/management/templates/edit_group.mako
```diff
@@ -4,138 +4,215 @@
...
+                        <span class="panel-entry">Discoverable: </span>
+                        <label>
+                        <input type="checkbox" name="new_discoverable" id="input_discoverable"
+                            %if discoverable:
+                               checked
+                            %endif
+                            %if group_name in MAGPIE_FIXED_GROUP_EDITS:
+                               disabled
+                            %else:
+                               onchange="document.getElementById('edit_discoverable').submit()"
```
dbyrns, line 136: Try to make a group discoverable [worked]. Then uncheck the box and got : 
400 Bad Request
Invalid POST request for group edit.
Is it a one way operation?

magpie/models.py
```diff
@@ -100,32 +110,107 @@ class ExternalIdentity(ExternalIdentityMixin, Base):
...
+
+        If request user is unauthenticated, :py:data:`magpie.constants.MAGPIE_LOGGED_USER` or itself, also grant
+        :py:data:`magpie.constants.MAGPIE_CONTEXT_PERMISSION` to allow access to contextually-available details.
+        (e.g.: user can view his own information and public ones)
+        """
+        user = self.request.user
+        acl = super(UserFactory, self).__acl__
+        if user and self.path_user and user.id == self.path_user.id:
+            acl += [(Allow, user.id, get_constant("MAGPIE_LOGGED_PERMISSION")),
+                    (Allow, user.id, get_constant("MAGPIE_CONTEXT_PERMISSION"))]
```
dbyrns, line 189: I look into permission.rst for more info on CONTEXT_PERMISSION, but I did not find it or anywhere else.
So what is the difference between logged info and contextual info?

magpie/ui/home/templates/template.mako
```diff
@@ -1,11 +1,20 @@
...
 <head>
-    <title>Magpie Administration</title>
+    <title>Magpie</title>
     <meta charset="UTF-8">
     <link rel="stylesheet" type="text/css" href="//fonts.googleapis.com/css?family=Open+Sans" />
-    <link rel="shortcut icon" type="image/x-icon" href="${request.static_url('magpie.ui.home:static/settings.png')}" />
-    <link href="${request.static_url('magpie.ui.home:static/style.css')}" rel="stylesheet" type="text/css" media="all" />
+    <link href="${request.static_url('magpie.ui.home:static/settings.png')}"
+          rel="shortcut icon" type="image/x-icon" />
+    <link href="${request.static_url('magpie.ui.home:static/themes/{}.css'.format(MAGPIE_UI_THEME))}"
```
dbyrns, line 9: You really provide a back-compatible theme! :)

magpie/ui/home/templates/error.mako
```diff
@@ -0,0 +1,44 @@
+<%inherit file="ui.home:templates/template.mako"/>
```
dbyrns, line 1: Are we supposed to get a magpie themed error page? Because when I saw the 400 invalid post to remove the customizable flag of a group I rather saw a vanilla html error page.

dbyrns (commented):
Review in progress (75%)

magpie/api/management/group/group_views.py
```diff
@@ -39,29 +42,52 @@ def get_group_view(request):
     """
     group = ar.get_group_matchdict_checked(request, group_name_key="group_name")
     return ax.valid_http(http_success=HTTPOk, detail=s.Group_GET_OkResponseSchema.description,
-                         content={u"group": gf.format_group(group, db_session=request.db)})
+                         content={"group": gf.format_group(group, db_session=request.db)})
 
 
-@s.GroupAPI.put(schema=s.Group_PUT_RequestSchema(), tags=[s.GroupsTag], response_schemas=s.Group_PUT_responses)
-@view_config(route_name=s.GroupAPI.name, request_method="PUT")
+@s.GroupAPI.patch(schema=s.Group_PATCH_RequestSchema(), tags=[s.GroupsTag], response_schemas=s.Group_PATCH_responses)
```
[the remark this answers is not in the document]
fmigneault: In which doc? README and generated OpenAPI but show that PUT are not available as replaced by PATCH.

magpie/alembic/versions/b739afcc91db_add_discoverable_groups.py
```diff
@@ -0,0 +1,25 @@
...
+# revision identifiers, used by Alembic.
+revision = "b739afcc91db"
+down_revision = "24da162a54f1"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column("groups", sa.Column("discoverable", sa.Boolean, nullable=False,
+                                      server_default=sa.schema.DefaultClause("0")))
```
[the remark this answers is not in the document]
fmigneault: Yes, by default at least so we don't expose something by mistake. 
There was no such thing as public groups, so its fair to assume they are not supposed to be.

magpie/ui/management/templates/edit_group.mako
fmigneault: Should be 2-way.
Probably the checkbox value is not correctly retrieved, (eg `None` instead of `False` for `discoverable`). 
Will look into it.

magpie/models.py
fmigneault: I agree, `MAGPIE_CONTEXT_PERMISSION` is not well detailed. 
It is a small extension of `MAGPIE_LOGGED_PERMISSION`. 

- MAGPIE_LOGGED_PERMISSION is when path-user == logged-user (but user is authenticated)
- MAGPIE_CONTEXT_PERMISSION is when path-user == logged-user OR path-user refers to unauthorized. 

Example is when unauthenticated user request public permissions using `/users/current/...` or when authenticated user requests public permissions with explicit `/users/anonymous/...`.

magpie/ui/home/templates/template.mako
fmigneault: Yes :D

magpie/ui/home/templates/error.mako
fmigneault: I've mostly tested for 401/403 use cases, but the intent is to wrap most of them. Will investigate.

magpie/alembic/versions/03b54feffe45_ensure_anonymous_group_applied_to_users.py
fmigneault: Locally, hard to say, but during Travis-CI execution, it always starts from scratch.

docs/configuration.rst
```diff
@@ -221,143 +332,225 @@ Following configuration parameters are used to define specific values that are r
...
+
+- | ``MAGPIE_LOGGED_PERMISSION`` [constant]
+  | (Value: ``"MAGPIE_LOGGED_USER"``)
+
+  .. versionadded:: 2.0.0
+
+  Defines a special condition of :term:`Access Permissions` related to the :term:`Logged User` session and the
+  targeted :term:`User` by the request. See details in :ref:`Route Access` for when it applies.
+
+- | ``MAGPIE_LOGGED_USER`` [constant]
```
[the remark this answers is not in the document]
fmigneault: It is constant (cannot override): 
https://github.com/Ouranosinc/Magpie/blob/user-scope-views/magpie/constants.py#L132

docs/configuration.rst
```diff
@@ -221,143 +332,225 @@ Following configuration parameters are used to define specific values that are r
...
-    targeted resource. In this situation, all *other* users would "lose" public permissions after they authenticate
-    themselves in `Magpie` as they would not be recognized as ``MAGPIE_ANONYMOUS_USER`` anymore.
   | (Default: ``${MAGPIE_ANONYMOUS_USER}``)
 
+  This parameter is preserved for backward compatibility of migration scripts and external libraries.
+  All users are automatically member of this :term:`Public` :term:`Group` to obtain :ref:`Inherited Permissions`.
+
+  .. warning::
+    To set :term:`Public` permissions, one should always set them on this :term:`Group` instead of directly on
+    ``MAGPIE_ANONYMOUS_USER`` as setting them directly on that :term:`User` will cause unexpected behaviours.
```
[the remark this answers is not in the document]
fmigneault: We could add an explicit check of the user when editing permissions to explicitly avoid that use case, only for sanity check/indicate group to target.

docs/configuration.rst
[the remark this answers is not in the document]
dbyrns: My bad, did not catch the constant aspect. But why include it in the configuration part of the doc?

magpie/alembic/versions/03b54feffe45_ensure_anonymous_group_applied_to_users.py
dbyrns: I agree backups have to be done and migrating bd is not without risks. At least tests cover the code that is modified. My fear was that you updated old upgrade fct to use new utility fct without possibility to test them. But obviously you had tests for that! :)

magpie/api/management/group/group_views.py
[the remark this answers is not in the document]
dbyrns: Just a vague memory, but I finally found what I has in mind here : https://github.com/Ouranosinc/Magpie/issues/349

>On the other hand, we could still support PUT *if an only if the full object details are provided in the body.

So the doc is ok! And we have an open issue to be backward compatible, it's fair enough for me. I don't think that these PUT were use a lot in the wild.

magpie/ui/management/templates/edit_group.mako
dbyrns: If you want to check the error I got (and for the error.mako template) you can check the group tb16 on our ogc-ades instance. I make it discoverable and it's stuck in this state.

magpie/models.py
dbyrns: Ok make sense. I tested it and found something, not sure if this is a bug with context_permission, but if I am not authenticated and try to access /users/(validuserid), I got a valid response... Expecting a not authorized or even a not found to not leak an actual user.

magpie/ui/home/static/style.css
```diff
@@ -54,41 +87,41 @@ input[type="submit"].button.delete {
...
 input[type="button"].button.warning:hover,
 input[type="button"].button.warning:focus,
 input[type="submit"].button.warning:hover,
 input[type="submit"].button.warning:focus {
     background-color: #FFAA33;
 }
 
-input[type="button"].img_button.cancel,
-input[type="submit"].img_button.cancel,
+input[type="button"].img-button.cancel,
```
dbyrns, line 123: Not so sure where exactly in the css the cancel button is affected but before it was green and now it looks disabled (should be blue according to the theme (or red)) and not even highlighted by the mouse.
image
Also it turns out that both admin and anonymous users seem to be editable which I guess should be prevented (maybe not related to this PR though)

magpie/ui/management/views.py
```diff
@@ -171,58 +188,62 @@ def view_users(self):
...
             if len(user_name) > get_constant("MAGPIE_USER_NAME_MAX_LENGTH"):
-                return_data[u"too_long_user_name"] = True
+                return_data["too_long_user_name"] = True
             if user_name in self.get_user_names():
-                return_data[u"conflict_user_name"] = True
+                return_data["conflict_user_name"] = True
             if user_name == "":
-                return_data[u"invalid_user_name"] = True
+                return_data["invalid_user_name"] = True
             if password is None or isinstance(password, six.string_types) and len(password) < 1:
```
dbyrns, line 229: Don't we have a parameter for the password length (rather than checking against < 1). Maybe that's why I got a json error when creating a user with a bad password rather than a smooth html invalid password warning.

dbyrns (changes requested):
Wow that was a good one!! Chrome even has a hard time to render that PR :) 
Overall this is a huge leap forward, great doc, many more tests, validation, error messages. Looks more and more like a product. 
As you can imagine, with such a big PR I found here and there some quirks, some could even be moved as issues and not prevent the merge of this PR. Just pinpoint the ones that require more time to fix and the related issue and I will gladly approve the PR once the other quick fixes are done.
Thanks!

docs/configuration.rst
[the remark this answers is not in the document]
fmigneault: Mostly for traceability of all provided values from `magpie.constants`. 
The docs is intended to be employed for both end-users and devs. 
If another dev has to work on Magpie, he can get more insight on what is the purpose of that constant.

docs/configuration.rst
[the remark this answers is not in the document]
fmigneault: #352

magpie/ui/home/templates/error.mako
fmigneault: Should all be wrapped and correctly redirectly now with default error decorator.

magpie/ui/management/templates/add_user.mako
```diff
@@ -8,70 +8,128 @@
...
+                <div class="alert-form-error"
+                    %if not (mismatch_password or invalid_password):
+                        style="visibility: hidden"
+                    %endif
+                >
+                    <img src="${request.static_url('magpie.ui.home:static/exclamation-circle.png')}"
+                         alt="ERROR" class="icon-error" />
+                    <div class="alert-form-text">
+                        %if invalid_password:
+                            Mismatch
```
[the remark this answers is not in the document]
fmigneault: resolved with e64159a, any bad-request received will be interpreted as best as possible

magpie/ui/management/templates/add_user.mako
```diff
@@ -8,70 +8,128 @@
...
+                        %elif mismatch_password:
+                            Invalid
```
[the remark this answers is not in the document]
fmigneault: also handled by e64159a fixes

magpie/ui/management/templates/add_user.mako
```diff
@@ -8,70 +8,128 @@
...
         </tr>
         <tr>
             <td>User group:</td>
-            <div class="input_container">
-                <td class="centered" colspan="2">
-                <select name="group_name" class="equal_width">
+            <td class="centered" colspan="2">
+                <div class="input-container">
+                <label>
+                <select name="group_name" class="equal-width">
```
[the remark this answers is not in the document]
fmigneault: adjusted first value to be `MAGPIE_ANONYMOUS_GROUP` in e64159a so that it is the selected one by default

magpie/ui/home/static/style.css
fmigneault: Regarding the edit, I will disable the buttons. The API does block the operation as expected, only not reflected on UI.

As for the 'cancel', the rendered colour is the "intended" one, but I agree it can lead to confusion with the 'disabled' one. I would like to avoid the red which is associated with destructive 'delete' operations. 
To better distinguish, I think the colour of 'disabled' should instead be modified for a faded/transparent-y version of its 'enabled' state. Something to imitate following scheme:

image

magpie/ui/management/views.py
fmigneault: e64159a now returns the explicit "Invalid 'password' ... (<12 characters)"  message

fmigneault:
@dbyrns 
> Thanks!

Thanks to you for the great feedback. You identified quite a lot that passed through the net.

magpie/models.py
fmigneault: Wow. That's weird. Specific tests are evaluating that and they pass... 
I will investigate, definitely not supposed to happen.

dbyrns: I don't know if you successfully reproduce that behavior. But if not, maybe it's related to the log out since before trying that I was logged and log out before retrying it.

magpie/ui/home/static/style.css
fmigneault: with b038ad8, this is the new rendering
also found that child resources where not very clear as not available when not allowed for given service/resource types

image

magpie/models.py
fmigneault: I think I identified the issue. Will had more explicit (and/or revise existing) test cases to validate this situation.

magpie/ui/home/static/style.css
dbyrns: After reviewing some UX sites, I agree that a gray cancel button is the way to go :P 
With your screenshot I just cannot see if moving the mouse over the button change its appearance (without highlighting, it can really appears disabled). Also, I would increase a little bit the constrast as shown on this site : https://uxmovement.com/buttons/why-cancel-buttons-should-never-have-a-color/ 
With that the cancel button would be perfect as your new disabled ones!

fmigneault: The combinations are as follow. Dark/light theme are controlled by Dark Reader mod which inverts colors or darkens all components. The contrasts are very subtle on the image (could be increased), but even that small change of color is very visible when passing from normal/hover.

image

dbyrns: Ok good, if the hovering is working with the new rendering (not the case with the actual demo instance) that's fine with me, the contrast don't bother me that much :P

fmigneault: > (not the case with the actual demo instance)

Really? I see the hover action on my side on that instance.

magpie/ui/management/templates/add_user.mako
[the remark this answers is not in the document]
dbyrns: Mismatch is now not handled. I create a new user with a valid password, type something else in the second box and it worked nevertheless

magpie/ui/management/views.py
dbyrns: The account self editing password functionnality returns a generic 500 error when setting a password less than 12 characters rather than showing an explicit message like with the new user form. (Same bad behavior when using the admin password editing)
image

magpie/ui/management/views.py
```diff
@@ -216,31 +218,59 @@ def add_user(self):
...
             if user_name in self.get_user_names():
-                return_data["conflict_user_name"] = True
+                return_data["invalid_user_name"] = True
+                return_data["reason_user_name"] = "Conflict"
             if user_name == "":
                 return_data["invalid_user_name"] = True
             if password is None or isinstance(password, six.string_types) and len(password) < 1:
                 return_data["invalid_password"] = True
-            elif password != confirm:
+            elif compare_digest(password, confirm):
```
dbyrns, line 238: Looks like this function is causing the password comparaison issue (weird)

dbyrns (changes requested):
Almost there, all my previous comments have been adressed, but I still saw 2 weird behaviors.

magpie/ui/management/views.py
fmigneault: Which python version are you running (version-dependent implementation)?

fmigneault: oh, yes, hmm... help to add `not` in front of that compare-equal
patch coming up

dbyrns (approved):
Thank you for all the fixes. We now have a very good 2.0 Magpie!

PR #353, opened 2020-09-25
Modified permissions scheme

### LINKED ISSUE #235: Support of DENY access
## Feature 

Consider the possibility to support the option to add an *non*-permission that would restrict the user access to a resource.
General use case: 
```
                            deny        allowed
service                                    X
├─ resource1               
│    ├─ resource2             X   
│    │    ├─ resource2.1  
│    │    └─ resource2.2                   X 
│    └─ resource3 
└─ resource4
```
Allowed user/group access: `service`, `resource1`, `resource3`, `resource4`
Denied user/group access: `resource2`, `resource2.1`, `resource2.2` (deny has priority)

## Changes
- This feature would need to refactor UI to have `Allow|Deny|None` instead of simple checkboxes.
- It would also require a new field during POST request to specify whether the created permission is `Allow|Deny`.
- Parsing of permissions would need to continue all the way to the top service to check for denial access in case lower allowed was given (current behaviour is to exit with allowed access directly).

### LINKED ISSUE #342: [Feature] Improve user/group permissions scheme
## Feature

When looking at permissions on user / group edit pages, a large matrix of checkboxes is listed.

With the eventual addition on 'deny' permission (#235), the list of applicable permissions can become quite large. 
(every combination of `deny`/`deny-match`, `<perm>`/`<perm>-match`)
For example, if a resource has `[read, write]` "permissions category" we would then have 8 "permissions variants" (using the previous scheme "match" which means non-recursive while, default is recursive) :

    read, read-match, write, write-match, deny-read, deny-read-match, deny-write, deny-write-match

This becomes worst the more permissions "category" we have. 

## API Changes

- Migrate all `[perm]-match`/`[perm]` permissions to respective `[perm]-allow-[match|recur]`
  (`[perm]` without modifier remains the `[recur]` variant for backward compatibility).
- Add new support for `[perm]-deny-[match|recur]` variants to existing sets of services.
- Add some resolution method that converts these applied permission variants to a effective permission 
    - saved permission in database must be a literal string for bw compat with ziggurat string permissions, which offers most of the auto-handling of relationships internally
    - since there would always be 3 parts (`[perm]-[allow|deny]-[match|recur]`), it will be easy to got from string to dict representation
- Code that handles new permissions must handle use-cases were there are conflicts 
  (how should we proceed with this??,  
   eg: 
   `[perm]-allow-match` applied when `[perm]-deny-match` already exists will not cause any problem from the db side as they are both unique strings. It wouldn't cause problem either when we resolve the effective permission (as deny has priority over allow). 
   This will cause problem in the sense that when `[perm]-allow-match` is applied, if one forgets to check for `[perm]-deny-match` already applied on the same resource, it will still appear as forbidden access although the user just added the allow permission. It is not very intuitive.
-  (!) __*edit*__: see decisions according to following thread comments

For each permission we would have the following properties :

    allow-match
    allow-recur
    deny-match
    deny-recur

The behavior should look like this given resource "/root/dir/file" :

``` python
bool authorized = False
if deny(file):
    return 401
if allow(file):
    authorized = True
# Parsing in bottom up order
parents = [dir, root]
for resource in parents:
    if deny-recursive(resource):
        return 401
    if allow-recursive(resource):
        authorized = True
if authorized:
    return 200
``` 

# UI Changes 

We should change the checkboxes to a combobox of all applicable permissions to make it more succinct. 

image

The currently existing checkbox view is still very relevant for overall viewing of *effective permissions*. 
This could be the visual rendering of results from #341
Because on that view, we will want only the 'allowed' vs 'denied' effective results, the way this result is obtained (recursive or match) does not matter. Checkboxes would be read-only in that case.

# Other

- Document all permission resolution methods (see also #332)
- While we are editing the permissions scheme, #334 could be done to add the HTTP-method-based permissions as synonyms for `read` and `write`.
- Add the UI helper tooltip for easier comprehension by user #335

fmigneault:
Fixes #235 
Fixes #342 

# Tasks

Summary of things listed in #342 that are being addressed: 

- [x] support permission modifiers
- [x] database migration to new permission format
- [x] documentation details about various permission modifiers
- [ ] documentation about *implicit*/*explicit* permission names, and between string/JSON representation.
- [x] modify POST request to handle conflicting permission *names* on resource regardless of applied modifiers
- [x] new DELETE request to remove a permission using JSON definition instead of name
- [x] new PUT request to replace or create the permission (to avoid needing to DELETE then POST each time)
- [x] handle ALLOW/DENY and RECURSIVE/MATCH modifiers during resource access effective resolution
- [x] change UI to have combobox/drop-down permission "modifiers" (with "blank" entry for no permission applied)
- [x] test(s) to validate that UI handles create/update/delete of new permissions scheme
   **update**: no extensive UI unittest, done manually while implementing
- [x] tests to validate expected resolution of ALLOW/DENY and RECURSIVE/MATCH 
- [x] tests to validate POST/PUT/DELETE operations on permissions, for every combination of user/group/service/resource, and for different situations (conflicts, no-conflict, edit modifier, other permission-name on same resource, etc.) 
- [x] add button UI button per-resource to show current resolution of effective permission for it.
- [x] ~make UI checkbox display effective permissions (disabled, read-only permissions resolved according to access/scope)~
  **update**: resolution per-resource computation is heavy with many items. Will require specific function that resolves them all at the same time to speed up. (moved to #362).
- [x] ~add `BROWSE`permission to `THREDDS`'s `Directory` to allow listing of *catalog* contents. 
    `READ` becomes relative only to `Files` (although `Directory` can have it for `RECURSIVE` resolution, but does not employ it by itself.~  
    **update**: moved to #361, see https://github.com/Ouranosinc/Magpie/pull/353#issuecomment-711079626


dbyrns (commented):
So far so good. Everything is now in place to implement the effective permission.
Do you plan to make the generic one up front or update the rest api implementation first?

magpie/permissions.py
```diff
@@ -33,29 +32,333 @@ class Permission(ExtendedEnum):
...
+        definition from literal string representation are parsed to generate the :class:`PermissionSet` instance.
+
+        :param permission: implicit or explicit permission name string, or any other known permission implementation
+        :raises ValueError: when the permission name cannot be identified or parsed
+        """
+        if isinstance(permission, PermissionSet):
+            return permission
+        if isinstance(permission, dict):
+            name = permission.get(
+                "name", permission.get("permission_name", permission.get("permission", permission.get("perm_name")))
```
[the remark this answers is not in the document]
fmigneault: `permission_name` and `permission` are for the old/new request body, and `perm_name` is the attribute of `ziggurat` objects.

fmigneault:
> So far so good. Everything is now in place to implement the effective permission.
> Do you plan to make the generic one up front or update the rest api implementation first?

I will implement the generic resolution under `ServiceInterface` so that all service types can employ it, but realistically will test it using `ServiceAPI` implementation.

magpie/permissions.py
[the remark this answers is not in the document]
dbyrns: I think it would be worth adding a comment with this information.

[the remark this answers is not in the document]
fmigneault: It is represented by the typing already.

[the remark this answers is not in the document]
dbyrns: I mean express that "permission_name" is for legacy reason while "permission" is the new request body. "perm_name" is the "ziggurat" attribute. I don't see that information anywhere else in this file. Maybe I'm missing something?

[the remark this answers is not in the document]
fmigneault: It is not the most common use-case. 
It is listed more a few lines below: 
https://github.com/Ouranosinc/Magpie/pull/353/files#diff-68582ecdeade7179085ce401678ddafcR287-R293

fmigneault:
@dbyrns (and maybe also @huard, @tlvu)

> Please check also for the thredds service as this is another tree based permissions scheme.

I'm at the point of implementing THREDDS's parsing methodology. 

At the moment, THREDDS allows [read, write] permissions.
It also allows child resource-types Directory and File, which both also have [read, write] permissions.
Directory can be nested any number of times, while File is a leaf (cannot nest other resource under it).

- Are these assumption valid? 
- Is there any other permission needed? 
- Is the nesting method valid?

Using these assumptions, the following uses cases would be possible and handled : 

- Does it make sense?

```
Legend::

    r: read     (interpreted as literal filesystem READ)
    w: write    (interpreted as literal filesystem WRITE)
    A: allow
    D: deny
    M: match
    R: recursive

Permissions Applied::
                                user            group           effective (reason/importance)
    Service1                    (w-D-M)         (w-A-R)         r-D, w-D  (user > group)
        Directory1              (r-A-R)                         r-A, w-A
            Directory2                          (r-D-R)         r-D, w-A  (revert user-res1)
                Directory3      (w-A-R)         (w-D-M)         r-D, w-A  (user > group)
                    File1                       (w-D-M)         r-D, w-D  (match > recursive)
        Directory4                              (r-A-R)         r-A, w-A
            File2               (w-D-M)                         r-A, w-D
            Directory5                                          r-A, w-A
                File3           (r-D-M)                         r-D, w-A
```

The next thing I need to know is how to convert an input request into the above structure. 
Currently, Magpie defines the conversion of the *request path* like follows, but I'm not sure how to make sense of it: 

https://github.com/Ouranosinc/Magpie/blob/4c9d011e28f2f2701c4ddecd8dd4ab01600c6178/magpie/services.py#L466-L479


- Is it that `fileServer`, `dodsC` and `catalog` are all the same "resource" accessed through different interfaces?
  (ie: does `/twitcher/ows/proxy/thredds/fileServer/[...]/file.x` and `/twitcher/ows/proxy/thredds/dodsC/[...]/file.x` represent the same file, in which case they share the same permissions, or actually 2 different resources with their own set of permissions).
- Are the permissions valid for each resource-type?  
   (I have seen a need for getting metadata, is it a specific permission distinct from READ/WRITE or simply READ permission onto a file *representing* metadata)
- Is there any other variants? (I have seen also `ncml`, `uddc`, and `iso`) Must they only be interpreted as different directories or there is something more to do with it?

dbyrns:
For the use cases shown, I think that everything make sense.
For the questions : 

> Is it that fileServer, dodsC and catalog are all the same "resource" accessed through different interfaces?

Yes, excactly! The same ressource can be accessed using different protocols and thus share the same permissions.

> Are the permissions valid for each resource-type?

@huard has already mentionned that it could be interesting to have a separate permission for browsing the catalog. Thus I would propose a "browse" permission mapped to the "catalog.html" element. @huard, would we require a fourth permission for metadata in ncml for example?

> Is there any other variants?

Since PAVICS, Ouranos have unlocked every access proposed by thredds and we now have : fileServer, dodsC, ncml, uddc, iso, wcs, wms. Which make me think that, if Francis P. confirm us that thredds wms is good enough, we will be able to drop ncwms support. (Pretty sure we will in fact).

huard:
ncml permissions should mirror dodsC permissions.

fmigneault:
@dbyrns 
I have played around a bit with the test THREDDS service provided here: 
https://remotetest.unidata.ucar.edu/thredds/dodsC/testdods/rtofs.nc.html

Using this, I came up will all possibilities I could come up with, listed here (tests all passing):
https://github.com/Ouranosinc/Magpie/pull/353/commits/ebdbb48c90db6cb05531735eac2b291287aa8770#diff-caee650c3003842e9277d696a063ab724d42606e64cbf44cf1deb9d1d67d1d98R369-R406

In the case of `catalog` specifically (either as `/thredds/catalog/dir` or `/thredds/[...]/catalog.html`, the "parent" Directory resource is employed to resolve user/group permissions. So setting `READ` with `MATCH` on that Directory resource would ONLY allow browsing the contents, but not fetching them. With `READ` and `RECURSIVE`, we get browsing + fetching, unless a `READ` & `DENY` was placed explicitly on the File resource under that Directory.

If anything else than `catalog`, it is the Dir/File path that is searched, which ignores the `dodsC`, `ncml`, etc. part, to consider all of the sub-parts after it.
So `/thredds/dodsC/someDir/someSubDir/someFile.nc.dds` will look for file `/someDir/someSubDir/someFile.nc`, with `dodsC` and `.dds` permute-able by basically anything. 

While testing, I was wondering if `WRITE` permission was ever used? 
The implementation *even before* this PR always returns enforced `READ` as requested permission, so I'm not even sure why `WRITE` is listed in applicable permissions.


+1 for dropping the `ncWMS2` service (its implementation on Magpie side is not that well defined anyway)

tlvu:
@fmigneault 

Here are just my feature feedback, not code feedback.

```
Permissions Applied::
                                user            group           effective (reason/importance)
    Service1                    (w-D-M)         (w-A-R)         r-D, w-D  (user > group)
        Directory1              (r-A-R)                         r-A, w-A
            Directory2                          (r-D-R)         r-D, w-A  (revert user-res1)
                Directory3      (w-A-R)         (w-D-M)         r-D, w-A  (user > group)
                    File1                       (w-D-M)         r-D, w-D  (match > recursive)
        Directory4                              (r-A-R)         r-A, w-A
```
The scenario above looks good.  Just to be sure, when user belongs to multiple groups, allow permission have precedence or deny?  Say `anonyous` and `users` groups have Deny but `projectA` group has Allow, Allow wins?

> * Is it that `fileServer`, `dodsC` and `catalog` are all the same "resource" accessed through different interfaces?
>   (ie: does `/twitcher/ows/proxy/thredds/fileServer/[...]/file.x` and `/twitcher/ows/proxy/thredds/dodsC/[...]/file.x` represent the same file, in which case they share the same permissions, or actually 2 different resources with their own set of permissions).
> 
> * Is there any other variants? (I have seen also `ncml`, `uddc`, and `iso`) Must they only be interpreted as different directories or there is something more to do with it?

I have touched on those questions here https://github.com/Ouranosinc/Magpie/issues/342#issuecomment-694235204

The list of services, their matching URL context root and which service is considered as Data or Metadata is listed.

The actual subtree text (ex: `dodsC` or `ncml` or `handle/future/sub-subtree`) should probably be configurable and not hardcoded to avoid changing Magpie code if Thredds config changes.  Someone may very well use `opendap` instead of `dodsC` for `serviceType="OpenDAP"` or not enable all those services.  Maybe create a new `thredds-service-config.yml` with content like:

```
thredds_services:
  data_type:
    - url_base: fileServer
    - url_base: dodsC
    - url_base: wcs
    - url_base: wms
  metadata_type:
    # no extra subpath, case when browsing using a browser
    - url_base: .  
    - url_base: ncml
    - url_base: uddc
    - url_base: iso
```

@huard 

> ncml permissions should mirror dodsC permissions.

I am not sure what you mean by this.

NCML link on `.nc` file sample: https://pavics.ouranos.ca/twitcher/ows/proxy/thredds/ncml/birdhouse/testdata/raven/ostrich-mohyse/Salmon-River-Near-Prince-George_meteo_daily.nc?catalog=https%3A%2F%2Fpavics.ouranos.ca%2Ftwitcher%2Fows%2Fproxy%2Fthredds%2Fcatalog%2Fbirdhouse%2Ftestdata%2Fraven%2Fostrich-mohyse%2Fcatalog.html&dataset=birdhouse%2Ftestdata%2Fraven%2Fostrich-mohyse%2FSalmon-River-Near-Prince-George_meteo_daily.nc

That's just XML so I would think this is metadata?

This is NCML link on a real `.ncml` file, still XML: https://pavics.ouranos.ca/twitcher/ows/proxy/thredds/ncml/datasets/simulations/bias_adjusted/cmip5/pcic/bccaqv2/day_bcc-csm1-1_historical+rcp85_r1i1p1_19500101-21001231_bccaqv2.ncml?catalog=https%3A%2F%2Fpavics.ouranos.ca%2Ftwitcher%2Fows%2Fproxy%2Fthredds%2Fcatalog%2Fdatasets%2Fsimulations%2Fbias_adjusted%2Fcmip5%2Fpcic%2Fbccaqv2%2Fcatalog.html&dataset=datasets%2Fsimulations%2Fbias_adjusted%2Fcmip5%2Fpcic%2Fbccaqv2%2Fday_bcc-csm1-1_historical%2Brcp85_r1i1p1_19500101-21001231_bccaqv2.ncml

I think we are mixing `.ncml` file vs NCML link type?

`.ncml` file are treated exactly like `.nc` file with same list of service type enabled, ex: https://pavics.ouranos.ca/twitcher/ows/proxy/thredds/catalog/datasets/simulations/bias_adjusted/cmip5/pcic/bccaqv2/catalog.html?dataset=datasets/simulations/bias_adjusted/cmip5/pcic/bccaqv2/day_bcc-csm1-1_historical%2Brcp85_r1i1p1_19500101-21001231_bccaqv2.ncml

dbyrns:
@fmigneault, I like the suggestion of @tlvu for metadata and data type configuration. read and browse permissions would be mapped on them and that would be future proof.

fmigneault:
@tlvu 
When there are multiple groups with permissions, DENY will have priority over ALLOW because there is not any way to tell which one is more "important" than the other.
@dbyrns 
I think I will go with a following PR for adding `BROWSE` + `[meta|data]_type` parameters parsed from config for `THREDDS` services.
This PR is already big enough, I would like to avoid even further extending its scope. 
For this type of dynamic config, I will need to adjust how Magpie takes in [providers.cfg](https://github.com/Ouranosinc/Magpie/blob/master/config/providers.cfg) and make sure it gets carried around down to where permission resolution occurs. 
For the moment, the config is dropped as soon as services registration was completed at startup.

dbyrns:
> I think I will go with a following PR for adding BROWSE + [meta|data]_type parameters

Yes that's fine. Let's close this one :)

tlvu:
> > I think I will go with a following PR for adding BROWSE + [meta|data]_type parameters
> 
> Yes that's fine. Let's close this one :)

Yes, good call, smaller PR easier to understand.

> When there are multiple groups with permissions, DENY will have priority over ALLOW because there is not any way to tell which one is more "important" than the other.

Agreed.  Just one clarification, multiple groups with permissions at the same level of precedence (recur vs match, match wins, recur from parent vs recur starting from child, recur starting from child, child wins) right?

fmigneault:
> Agreed. Just one clarification, multiple groups with permissions at the same level of precedence (recur vs match, match wins, recur from parent vs recur starting from child, recur starting from child, child wins) right?

Exactly.

fmigneault:
@dbyrns 
`3.0.0-rc` is deployed on https://ogc-ades.crim.ca/magpie for review

magpie/services.py
```diff
@@ -72,85 +65,315 @@ def __init__(self, service, request):
...
+            for perm_name in requested_perms:
+                for perm_tup in cur_res_perms:
+                    perm_set = PermissionSet(perm_tup)
+
+                    # if user is owner (directly or via groups), all permissions are set,
+                    # but continue processing this resource until end in case user explicit deny reverts it
+                    if perm_tup.perm_name == ALL_PERMISSIONS:
+                        for perm in requested_perms:
+                            all_perm = PermissionSet(perm, perm_set.access, perm.scope, perm.type)
+                            if perm_set.access == Access.DENY:
```
dbyrns, line 290: I don't want to be picky, but this function is the most interesting part of this PR.
I would like to have a little bit more comment on this if/else.
The if is about a user that own a ressource but is denied access to it?

magpie/services.py
```diff
@@ -72,85 +65,315 @@ def __init__(self, service, request):
...
+                                effective_perms[perm] = all_perm
+                            else:
+                                effective_perms.setdefault(perm, all_perm)
```
dbyrns, line 293: ...and here we allow the permission unless, already denied in the dict?

magpie/services.py
```diff
@@ -72,85 +65,315 @@ def __init__(self, service, request):
...
+                    elif perm_set.name not in requested_perms or perm_set.name != perm_name:
+                        continue
+                    # only first resource can use match, parents are recursive-only
+                    if not match and perm_set.scope == Scope.MATCH:
+                        continue
+
+                    # user direct permissions have priority over inherited ones from groups
+                    if perm_set.type == PermissionType.DIRECT:
+                        perm = effective_perms.get(perm_name)
+                        # explicit user deny overrides user allow if any already found
```
dbyrns, line 304: The if seems to also include the case where we have an explicit allow that also override inherited permissions (whatever the type). Should also be in the comment. I know that previous comment indicate that. But given the complexity of the algo, I think that the comment should match exactly what the if does.

magpie/services.py
```diff
@@ -72,85 +65,315 @@ def __init__(self, service, request):
...
+                        if perm is None or perm.type == PermissionType.INHERITED or perm_set.access == Access.DENY:
+                            effective_perms[perm_name] = perm_set
+                        continue  # final decision for this user, skip any group permissions
+
+                    # otherwise check if for another group permission and yet no user permission
+                    perm = effective_perms.get(perm_name)
+                    if perm is None or perm.type == PermissionType.INHERITED:
+                        # explicit deny overrides allow if any already found
+                        if perm is None or perm_set.access == Access.DENY:
```
dbyrns, line 313: here, if I correctly interpret the ifs it should resolve to this:
if perm is None or (perm.type == PermissionType.INHERITED and perm_set.access == Access.DENY):
So here we set the permission if not already set : fine
or if the current permission is inherited and we got a deny : ok 
Since there is only 2 cases where the perm is set, do you think that the ifs could be simplified?

dbyrns (commented):
Work in progress.

magpie/services.py
fmigneault: That would be the idea, but it is not really doing anything at the moment since there is no actual owner currently. 
I can add a FIXME note for when/if ownership rules get implemented properly.

magpie/services.py
fmigneault: Yes. It looks this way because of the above block that had a similar 2-tier ifs, but I don't mind. The interpretation is like you resolved it.

fmigneault:
@dbyrns adjusted with comment details as per review

dbyrns:
Regarding ea2455d, you were not supposed to start another PR for that?

magpie/services.py
```diff
@@ -298,20 +304,25 @@ def effective_permissions(self, user, resource, permissions=None, allow_match=Tr
...
-                    # otherwise check if for another group permission and yet no user permission
+                    # otherwise check for group permission
+                    # like previously, explicit DENY overrides ALLOW if permission name was already found
+                    # if permission name not already found, ALLOW/DENY is set regardless (first occurrence)
                     perm = effective_perms.get(perm_name)
-                    if perm is None or perm.type == PermissionType.INHERITED:
-                        # explicit deny overrides allow if any already found
-                        if perm is None or perm_set.access == Access.DENY:
-                            effective_perms[perm_name] = perm_set
+                    if perm is None or (perm.type == PermissionType.INHERITED and perm_set.access == Access.DENY):
```
dbyrns, line 324: Thanks for the update, much easier to follow. I think it still miss a point about the perm.type == PermissionType.INHERITED.
If the perm is direct (line 312), an allow permission overrides any access type that has been inherited. It means that a direct "allow" takes precedance over inherited "deny",
Here, to override any inherited permissions the access type must be deny. It means that a "deny" group permission take precedance over other "allow" group permission.
In short, we should better explain the difference between INHERITED  or DENY of line 316 vs INHERITED and DENY of line 324

fmigneault: I feel that is what comments on lines 311 and 318 describe,

fmigneault:
> 
> 
> Regarding [ea2455d](https://github.com/Ouranosinc/Magpie/commit/ea2455d9cf233d0d58846743a7078f5b4dbebf59), you were not supposed to start another PR for that?

Yes, look at the branch name [permission-extensions](https://github.com/Ouranosinc/Magpie/compare/permission-extensions)

This one is [permissions-scheme](https://github.com/Ouranosinc/Magpie/tree/permissions-scheme)

magpie/services.py
dbyrns: I'll try to be more explicit. The following 3 rules should be clear while reading the comments:
1. direct > inherited (line 311 is fine and line 318 cover the fact that once direct is set we're done, but line 314 put emphasis on DENY override ALLOW while opposite is true too (the reason why we got a type == inherited **or** access == DENY is not covered) 
2. direct DENY > direct ALLOW  (line 314)
3. inherited DENY > inherited ALLOW (line 321 is confusing because it's stating **like previously** while this time it's between group (the reason why we got a type == inherited **and** access == DENY is not covered)

magpie/ui/management/views.py
```diff
@@ -329,14 +330,17 @@ def edit_user(self):
...
-            elif "edit_permissions" in self.request.POST:
-                if not res_id or res_id == "None":
-                    remote_id = int(self.request.POST.get("remote_id"))
-                    services_names = [s["service_name"] for s in services.values()]
-                    res_id = self.add_remote_resource(cur_svc_type, services_names, user_name, remote_id, is_user=True)
-
-                removed_perms, new_perms = \
-                    self.edit_user_or_group_resource_permissions(user_name, res_id, is_user=True)
+            elif "edit_permissions" in self.request.POST and not inherit_grp_perms:
+                # FIXME:
```
dbyrns, line 334: Why doing stuff in batch remove the need to add remote ressources?
If I want to add a permission on something that doesn't already exist, it seems like the add_remote fct should be required, no?

dbyrns (commented):
Good job with all that. The tests are pretty exhaustives!
Here is some issues I got with the UI:

1. Cannot collapse tree using the small arrow (the text do the job but it's less obvious)
2. The apply button position isn't appropriate for me, it's goal is not clear at first. Renamed to "Apply permissions" and moved to the bottom right under the tree view along with a Cancel  button that simply refresh the page could do the trick, don't you think?
3. I can leave the page with pending changes without any notice. Using https://stackoverflow.com/questions/7317273/warn-user-before-leaving-web-page-with-unsaved-changes and the dirty flag could solve that and even improve issue #2 solution by enabling the button only when changes are done.
4. Why the effective permission test button is only enabled with inherited permission checked? Is it because the effective resolution doesn't exist for the current user only? Maybe we should check the view inherited group permissions checkbox by default so that we can see this nice feature up front?

dbyrns:
I'm also wondering if I miss something with the test effective permission button... In this screenshot, given that I have a direct deny recursive permission, should I got uncheck marks for the tested ressources?
image

fmigneault:
> 1. Cannot collapse tree using the small arrow (the text do the job but it's less obvious)

Yes. The event is only on the line-text. I struggled a bit to make it work because adding the event directly on the list-item was firing when clicking on the selects. I defined a "hover-area" using the text only to avoid it. I can try looking into adding the event only on the `<li> ::marker` as well. I guess it should be possible. 

> 2. The apply button position isn't appropriate for me, it's goal is not clear at first. Renamed to "Apply permissions" and moved to the bottom right under the tree view along with a Cancel  button that simply refresh the page could do the trick, don't you think?
> 3. I can leave the page with pending changes without any notice. Using https://stackoverflow.com/questions/7317273/warn-user-before-leaving-web-page-with-unsaved-changes and the dirty flag could solve that and even improve issue #2 solution by enabling the button only when changes are done.

I agree with renaming it to make it more explicit. The reason I put the button on top was purposely to avoid the situation of changing permissions and forgetting to press apply at the bottom without going into the full dirty form handling.

> 4. Why the effective permission test button is only enabled with inherited permission checked? Is it because the effective resolution doesn't exist for the current user only? Maybe we should check the view inherited group permissions checkbox by default so that we can see this nice feature up front?

Yes. Effective are by definition user+inherited. When testing, I found that it was really confusing to have only user permissions displayed and have things allowed/denied without the corresponding checkboxes and permissions displayed in the selects.

I find I go into that view way more often to set permissions than to test effective ones. I also find viewing user+group permissions by default is not natural since we arrive there from clicking "Users" button. We would be greeted with a page full of inactive options, and it is not intuitive to know we need to click the checkbox to make them active.

fmigneault:
@dbyrns Regarding the permission, is your user an admin? If so, you got full access for effective permissions.

Using non-admin user: 
image

magpie/services.py
fmigneault: Maybe it is because of the way I parse the sequence that doesn't feel ambiguous for me. 

1. User > Group - so if Group was set (ALLOW or DENY) we don't care, we set whatever User permission says. 
   For this reason, the opposite use case is never true.
2. We then enter User branch. We are in the realm of User-only permissions. So DENY > ALLOW is the only thing to evaluate.
----
3. ONLY THEN, once we established we are not evaluating User permissions, we compare only Group with Group. 
4. Again, the only thing to check is DENY > ALLOW resolution. 

It is mostly for this reason I had 2 ifs before. It gave a clean distinction between the 2 phases (eval User > Group) and then, for each (DENY > ALLOW). 
If you think about it, the *original* 2nd if of the Group block is identical to the 2nd one of the User group 
`if perm is None or perm.type == PermissionType.INHERITED or perm_set.access == Access.DENY:`
Only thing is that we have a redundant check for `perm.type == PermissionType.INHERITED` since we evaluated that on the 1st if of the Group block, which is why it became `if perm is None or perm_set.access == Access.DENY:`

Making the if all-in-one for Group portion makes it more complicated to evaluate in my opinion because it introduces the mixing of Type and Access simultaneously.

dbyrns:
> @dbyrns Regarding the permission, is your user an admin? If so, you got full access for effective permissions.

Haha, that's it! Everything works fine.

dbyrns:
Regarding my 4 issues : 
1. If you can cover the li that would be cool (the arrow was working before, don't know how however). But if this is a big job, never mind, admins will get used to it.
2. Ok for renaming. Do you think that moving it to the top-right at least (nearer to the permissions) would be a good compromise?
3. I did it for a small project and it is really just setting 3 js listeners on document load: combo updated (set dirty), cancel/apply (reset dirty), unloaded (if dirty, warn). It is a nice to have tough and useless if we move to asynch js, so forget it, that was just FYI.
4. You're right, again :) I do think, however, that we could rename the checkbox for "View group inherited **and effective** permissions" to get a clue about the nice effective stuff. What do you think?

magpie/ui/management/views.py
dbyrns: This is because david c. did that part, but the trick is that the ui shows ressources that doesn't exist in magpie db. It is using the synched remote ressources db (if I remember correctly) and mix it with the magpie ressources. 
So the synch doesn't really populate the magpie ressources, but the remote db and the mix occurs in real time for the ui. So checking a ressource doesn't mean that it already exists. If that ressource came from the remote db it must first be added to the magpie db before setting its permission.

magpie/services.py
dbyrns: > 1. [...] For this reason, the opposite use case is never true.

I meant if group was DENY, but user is ALLOW, allow override deny. Just explaining that would explain perfectly why we have a inherited or deny. For me line 311, 314 and 318 doesn't make that obvious.

> 2. [...] So DENY > ALLOW is the only thing to evaluate. 

This statement cover the second part of the if of l316. And the comment just above cover only that part. My point is reallly to cover the inherited part too. Comment should look like that : "permission is set if not already found (first occurence), if inherited regardless of access (direct > inherited) or deny (deny > allow)"

> 3 [...] we compare only Group with Group

Exact. But comment said "like previously" which compare direct with group and direct with direct.
I would change the comment to put the emphasis on the group with group **and** deny > allow to reflect the if, something like : "otherwise check for group permission, permission is set if not already found (first occurence) or if inherited (compare group with group only) AND deny (deny > allow)"

Sorry to be bothering, but I'm only looking to have better comments explaining what is going on for non-initiated, not to refactor the whole part.

fmigneault:
>     1. If you can cover the li that would be cool (the arrow was working before, don't know how however). But if this is a big job, never mind, admins will get used to it.

I found a way to do it using a fake div area. I couldn't apply the event on the marker directly. 
It is deployed on ogc-ades, using 3.1.0-rc so #363 is also included. 


>     2. Ok for renaming. Do you think that moving it to the top-right at least (nearer to the permissions) would be a good compromise?

It doesn't align very nicely with `Sync` button when it is there. 
Will need to flip the elements around a bit. (not hard, just didn't notice this message update until after deployed)

>     4. You're right, again :) I do think, however, that we could rename the checkbox for "View group inherited **and effective** permissions" to get a clue about the nice effective stuff. What do you think?

Sounds good.

magpie/ui/management/views.py
fmigneault: We will need to revisit this whole functionality IMHO. It is not very well maintained. 
For example, we cannot update the service type at later date using the API, and creating a service via the UI doesn't provide any selector to specify which one we want.
I have also booted a quick demo THREDDS server and it fails to sync files (with force sync button) due to internal parsing.

dbyrns:
I still see 3 pending issues : 
1. https://github.com/Ouranosinc/Magpie/pull/353#discussion_r512742704
2. https://github.com/Ouranosinc/Magpie/pull/353#issuecomment-717698950, case #2
3. https://github.com/Ouranosinc/Magpie/pull/353#issuecomment-717698950, case #4

fmigneault:
> 
> 
> I still see 3 pending issues :
> 
>     1. [#353 (comment)](https://github.com/Ouranosinc/Magpie/pull/353#discussion_r512742704)
> 
>     2. [#353 (comment)](https://github.com/Ouranosinc/Magpie/pull/353#issuecomment-717698950), case #2
> 
>     3. [#353 (comment)](https://github.com/Ouranosinc/Magpie/pull/353#issuecomment-717698950), case #4

 1. addressed via 2727e24

 2. & 3. addressed via 1322770  of PR #363

dbyrns (approved):
Well done! Another great addition to Magpie.

PR #363, opened 2020-10-23
Permission extensions

### LINKED ISSUE #157: Potential incorrect WPS-1 service ACL permissions validation
see crim-ca/weaver#22 for use case

- [ ] Verify that `POST` for WPS-1 `request=Execute` doesn't cause problem (when using XML body instead of `DataInputs=...` in URI with `GET` request). When request has XML body, the values inside it are used instead of `identifier=...&request=...` in the URI. Therefore, the `ServiceWPS` possibly fails to retrieve the actual  `request` field, and then fails to retrieve the proper ACL for Twitcher.
- [ ] verify that WPS-1 `request=...` to corresponding Magpie service permission mapping is properly accomplished so that access to corresponding requests is granted/blocked as expected.

### LINKED ISSUE #234: WMS parsing of THREDDS service
Support the new functionality of integrated WMS in Thredds.

### LINKED ISSUE #361: [Feature] BROWSE permission for ServiceTHREDDS
- [x] add `BROWSE`permission to `THREDDS`'s `Directory` to allow listing of *catalog* contents. 
    `READ` becomes relative only to `Files` (although `Directory` can have it for `RECURSIVE` resolution, but does not employ it by itself. 
- [x] update doc about new behaviour

see details: 
- https://github.com/Ouranosinc/Magpie/pull/353#issuecomment-711079626
- https://github.com/Ouranosinc/Magpie/pull/353#issuecomment-710495259

fmigneault:
Depends on #353 (rebased). 

- Adds additional `BROWSE` permission to `THEDDS` service (#361)
- Adds many doc details about service implementations.
- Fix and adds test cases for #157 
- Fixes and test some edge case that failed to grant access when target `Resource` was missing (not explicitly defined in db), but that parent `Resource` had recursive-scope `Permission`.

Fixes #157 
Fixes #234
Fixes #361

### DIFF
--- a/docs/services.rst
+++ b/docs/services.rst
@@ -407,6 +407,8 @@ When using the `Magpie` Docker image, the default command run the `magpie-cron`_
 cron job will periodically execute the :term:`Resource` auto-synchronization feature for a given :term:`Service` that
 supports it.
 
+The synchronization mechanism can be launched from `Magpie` UI using the ``Sync`` button located on relevant pages.
+
 .. seealso::
 
     Utility ``magpie_sync_resources`` in :ref:`utilities_helpers` is also available to manually launch a
--- a/docs/services.rst
+++ b/docs/services.rst
@@ -1,9 +1,6 @@
 .. _services:
 .. include:: references.rst
 
-.. employ the permissions as base for this chapter to ease reading by shorter reference definitions
-.. py:currentmodule:: magpie.permissions
-
 ===========
 Services
 ===========
@@ -25,7 +22,7 @@ Therefore, they can be listed and searched for either using ``/services`` API ro
 On top of any :term:`Resource`'s metadata, a :term:`Service` provides specific information about its location, its
 remote synchronization method (if any), and its exposed endpoint. Another important detail about the :term:`Service`
 is its ``type``. This will not only dictate its purpose, but also define the whole schema of allowed :term:`Resource`
-under it (if any), as well as every one of their :ref:`Allowed Permissions`.
+under it (if any), as well as every one of their :term:`Allowed Permissions`.
 
 The final distinction between a :term:`Service` and generic :term:`Resource` is their position in the hierarchy. Only
 :term:`Service`-specialized :term:`Resource` (literally ``resource_type = "service"``) are allowed to be placed at the
@@ -89,6 +86,9 @@ On top of the above methods, the following attributes must be defined.
 Available Services
 -------------------
 
+.. employ the permissions as base for this section to ease reading by shorter reference definitions
+.. py:currentmodule:: magpie.permissions
+
 .. seealso::
     Module :py:mod:`magpie.services` contains the implementation of every :term:`Service` presented below.
 
@@ -148,7 +148,7 @@ each sub-:term:`Resource` as presented below.
 
 Every :class:`magpie.models.Route` as well as the :term:`Service` itself can have the :term:`Permission` based on the
 HTTP method of the incoming request. All :class:`Access` and :class:`Scope` modifiers are also supported for highly
-customizable :term:`ACL` combinations. See `permission_modifiers`_ for further details.
+customizable :term:`ACL` combinations. See :ref:`permission_modifiers` for further details.
 
 ServiceTHREDDS
 ~~~~~~~~~~~~~~~~~~~~~
--- a/docs/services.rst
+++ b/docs/services.rst
@@ -204,7 +204,7 @@ and the file *extension*. The *default* methodology employed to categorize these
 by the below configuration.
 
 .. note::
-    A custom categorization between *metadata*/*data* contents can be provided With either the `providers.cfg`_ or
+    A custom categorization between *metadata* and *data* contents can be provided With either the `providers.cfg`_ or
     a :ref:`config_file` as described in greater lengths within the :ref:`configuration_link` chapter.
 
 .. code-block:: YAML
@@ -221,7 +221,7 @@ by the below configuration.
           - .*.nc
         metadata_type:
           prefixes:
-            - null  # note: special value evaluated as `no-prefix`, use quotes to handle special keywords if needed
+            - null  # note: special YAML value evaluated as `no-prefix`, use quotes if literal value is needed
             - catalog
             - ncml
             - uddc
@@ -264,13 +264,15 @@ used to lookup the :term:`Resource`. Otherwise, the plain ``<file>`` name is use
 if ``file_patterns`` is specified as empty or ``null``. Not explicitly overriding the field will result into using the
 above *default* ``file_patterns``. The ``file_patterns`` allow for example to consider ``file.nc``, ``file.ncml`` and
 ``file.nc.html`` as the same :term:`Resource` internally, which avoids duplicating :term:`Applied Permissions` across
-multiple :term:`Resource` for every *metadata*/*data* representation.
+multiple :term:`Resource` for every *metadata* or *data* representation.
 
 ServiceBaseWMS
 ~~~~~~~~~~~~~~~~~~~~~
 
 .. seealso::
+
     Derived implementations:
+
     - `ServiceGeoserverWMS`_
     - `ServiceNCWMS2`_
 
@@ -294,6 +296,7 @@ ServiceGeoserverWMS
 ~~~~~~~~~~~~~~~~~~~~~
 
 .. seealso::
+
     Base class: `ServiceBaseWMS`_
 
 This implementation is defined by :class:`magpie.services.ServiceGeoserverWMS`. It extends the base class by using
@@ -311,6 +314,7 @@ ServiceNCWMS2
 ~~~~~~~~~~~~~~~~~~~~~
 
 .. seealso::
+
     Base class: `ServiceBaseWMS`_
 
 This implementation is defined by :class:`magpie.services.ServiceNCWMS2`. It extends the base class by using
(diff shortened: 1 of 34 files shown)


docs/services.rst
```diff
@@ -15,78 +19,384 @@ Each :term:`Service` is a specialization of a :term:`Resource` (see: :class:`mag
...
 Service Synchronization
 ------------------------
 
-.. todo:
-    resource auto-sync feature for a given service and cron configuration setup for it
+.. versionadded:: 0.7
+
+Some :term:`Service` implementations offer a synchronization feature which is represented by field ``sync_type`` within
+`Magpie` API responses. When this option is available, the corresponding :term:`Service` is able to query the real
+*remote service provider* to retrieve actual :term:`Resource` nested under it. This allows quick generation of the full
```
dbyrns, line 384: I think that the tree hierarchy is not fully recovered... only at a given depth.

docs/services.rst
```diff
@@ -15,78 +19,384 @@ Each :term:`Service` is a specialization of a :term:`Resource` (see: :class:`mag
...
+
+When using the `Magpie` Docker image, the default command run the `magpie-cron`_ utility in parallel to the API. This
+cron job will periodically execute the :term:`Resource` auto-synchronization feature for a given :term:`Service` that
+supports it.
 
 .. seealso::
 
-    Utility ``magpie_sync_resources`` in `Magpie CLI Helpers`_ is also available to manually launch a :term:`Resource`
-    synchronization operation for supporting :term:`Service`-types.
+    Utility ``magpie_sync_resources`` in :ref:`utilities_helpers` is also available to manually launch a
```
dbyrns, line 400: Maybe refer also to the ui synch button...

magpie/services.py
```diff
@@ -672,48 +681,64 @@ class ServiceTHREDDS(ServiceInterface):
...
+    def get_config(self):
+        # type: () -> ConfigDict
+        cfg = super(ServiceTHREDDS, self).get_config()
+        cfg.setdefault("file_patterns", [r".*.nc"])
+        cfg.setdefault("data_type", {"prefixes": []})
+        if not cfg["data_type"]["prefixes"]:
+            cfg["data_type"]["prefixes"] = ["fileServer", "dodsC", "dap4", "wcs", "wms"]
+        cfg.setdefault("metadata_type", {"prefixes": []})
+        if not cfg["metadata_type"]["prefixes"]:
+            cfg["metadata_type"]["prefixes"] = [None, "catalog", "ncml", "uddc", "iso"]
```
dbyrns, line 693: You provide a great default configuration, but to avoid a back-compatibility migration we could also push all prefixes to the data type. The responsability of the admin overriding this config with metadata prefixes would then have to set the permissions of his instance accordingly.
Note that I would prefer a proper migration, but understand that it could be complex.

magpie/services.py
```diff
@@ -672,48 +681,64 @@ class ServiceTHREDDS(ServiceInterface):
...
-        while found_resource and path_parts:
+        found_resource = child_resource = self.service
+        while child_resource and path_parts:
             part_name = path_parts.pop(0)
-            found_res_id = found_resource.resource_id
-            found_resource = models.find_children_by_name(part_name, parent_id=found_res_id, db_session=self.request.db)
+            # when reaching the final part, test for possible file pattern, otherwise default to literal value
+            #   allows combining different naming formats into a common file resource (eg: extra extensions)
+            # if final part is a directory, still works because of literal value
+            #   directory name much match exactly, no format naming variants allowed
```
dbyrns, line 712: typo "much" -> "must"

dbyrns (changes requested):
Good job! I only got some small questions and suggestions.

docs/services.rst
```diff
@@ -15,78 +19,384 @@ Each :term:`Service` is a specialization of a :term:`Resource` (see: :class:`mag
...
+The implementation of this :term:`Service` is handled by class :class:`magpie.services.ServiceAccess`. It is intended to
+act as a simple *all-or-nothing* endpoint barrier, where only :attr:`Permission.ACCESS` can be applied, and only
+directly on the :term:`Service` itself. A :term:`User` or :term:`Group` that has that :term:`Permission` with
+:attr:`Access.ALLOW` will be able to reach the :term:`Service`. Any other operation (or explicit :attr:`Access.DENY`)
+will all result into denied access to the private URL registered by the :term:`Service`.
+
+.. versionchanged:: 3.0
+    This :term:`Service` implementation dates prior to the integration of :class:`Access` and :class:`Scope` concepts
+    that now provides modifiers to the :term:`Permission` resolution methodology, notably prior to the addition of
+    :attr:`Access.DENY`. This :term:`Service` is equivalent to a simplified version of the fine-grained `ServiceAPI`_
```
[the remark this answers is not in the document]
fmigneault: Could be. I leave it for now because it is not really painful to maintain.

docs/services.rst
```diff
@@ -15,78 +19,384 @@ Each :term:`Service` is a specialization of a :term:`Resource` (see: :class:`mag
...
+    The :attr:`Permission.BROWSE` permission is used to provide listing access of contents when targeting a
+    :term:`Resource` of type :class:`magpie.models.Directory`. When targeting a :class:`magpie.models.File`, it instead
+    provides *metadata* access to that file.
+
+Permission :attr:`Permission.READ` can be applied to all of the resources, but will only effectively make sense when
+attempting access of a specific :term:`Resource of type :class:`magpie.models.File`.
+
+.. versionchanged:: 3.1
+    Permission :attr:`Permission.READ` does not offer *metadata* content listing of :class:`magpie.models.Directory`
+    anymore. For this, :attr:`Permission.BROWSE` should be used instead. Setting :attr:`Permission.READ` on a
```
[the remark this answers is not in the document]
fmigneault: Nothing is done for the moment, but we could had this step to the db migration scripts.
That's a good point actually. Users can only download but not query metadata at the moment if READ only.

docs/services.rst
```diff
@@ -15,78 +19,384 @@ Each :term:`Service` is a specialization of a :term:`Resource` (see: :class:`mag
...
+:term:`Effective Permissions` resolution, the value of the query parameter ``identifier`` is employ to attempt mapping
+it against an existing :term:`Resource`. The resolution of :term:`Effective Permissions` in even of multi-level tree
+:term:`Resource` is computed in the usual manner described in the :ref:`Permissions` chapter.
+
 
+.. warning::
+    When applying permissions on a per-:class:`magpie.models.Process` basis, :attr:`Scope.MATCH` modifier is recommended
+    if only a specific ``request`` should be granted access, although :attr:`Scope.RECURSIVE` would have the same effect
+    currently because these resources are necessarily leaf nodes by definition. This is to prevent unexpectedly granting
+    additional lower-level :term:`Resource` access in the event this definition gets modified or extended in the future.
```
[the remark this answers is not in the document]
fmigneault: The Service and the children Resource can have very distinct permission names if we want. 
I actually use this for `ServiceWPS` vs `Process` under it. Only `ServiceWPS` can have `GET_CAPABILITIES`, and both can have `DESCRIBE_PROCESS` and `EXECUTE`.

docs/services.rst
fmigneault: Specifically for THREDDS, there is an hardcoded `max_depth = 3`. Others return `None` which will fetch everything. 
Good catch.

docs/services.rst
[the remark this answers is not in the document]
dbyrns: Hum, that's good! You just cannot restrict MATCH over RECURSIVE to avoid the potential issue listed above.

magpie/services.py
fmigneault: Indeed. There is quite a bit of work to move this to the db properly. 
Saving the config the db in itself is somewhat easy, but the problem is about how to handle updating it. 

Example: 
First time the config is provided via config.yml and stored to database. 
Then, if omitted another time, how do we handle this? Reuse the thing from db or flush? How to know what is the intended behavior? The only real method would be to have an API call to update, but then more development to support this.

For now, I think it is safe to assume that the config.yml provided will always be there, so the config should be loaded each time as expected.

dbyrns: Hum I'm not sure you understand. I don't want to save the config to the db.
I only point out that for an instance without proper thredds configuration (all of them as of speaking), when we will update magpie, the ``if not cfg[...][...]`` will be true and the "default" behavior will apply. This default behavior, with the current instances, will cause to remove all permissions for the metadata_type [None, "catalog", "ncml", "uddc", "iso"] since no BROWSE permissions exist. 
What I propose here, is a solution to https://github.com/Ouranosinc/Magpie/pull/363#discussion_r512801739. If by default, you put all the prefix in ``cfg["data_type"]["prefixes"]``, then when upgrading magpie, even without BROWSE permissions set, everything will work as before. 
The admin will then be free to configure the browse permission for its users/groups and later configure thredds properly using the config.yml.

fmigneault: Although that would work, it somewhat goes against to point of adding the permission specifically for metadata in the first place. Isn't it be better to have a default config that matches as closely as possible "normal" usage? I would argue that if for whatever reason we want to preserve the old method, we should provide the config to do so.

When Magpie is going to be updated to support this specific use case, proper configuration should be provided to it.
I think it is cleaner to duplicate existing READ entries as equivalent BROWSE during migration.

dbyrns: Yes I agree, that's why I noted that I would prefer a proper migration. I just tought that it may be hard, but let's do that. Assign BROWSE permissions where READ permissions currently exist.

docs/services.rst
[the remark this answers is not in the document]
fmigneault: added by d364884

docs/services.rst
```diff
@@ -15,78 +19,384 @@ Each :term:`Service` is a specialization of a :term:`Resource` (see: :class:`mag
...
+permission will be immediately assumed as :attr:`Access.DENY` regardless of type. To allow top-level access directly on
+the :term:`Service`'s root without ``<prefix>``, it is important to provide ``null`` within the desired ``prefixes``
+list. Duplicates between the two lists of ``prefixes`` will favor entries in ``metadata_type`` over ``data_type``.
+
+After resolution of the content type from ``<prefix>``, the resolution of any amount of :class:`magpie.models.Directory`
+:term:`Resource` will be attempted. Any missing children directory :term:`Resource` will terminate the lookup process
+immediately, and :term:`ACL` will be resolved considering :attr:`Scope.RECURSIVE` if any applicable parent
+:term:`Resources` for the given :term:`Permission` selected by ``<prefix>`` and from where lookup stopped.
+
+Once the last element of the path is reached, the ``file_patterns`` will be applied against ``<file>`` in order to
```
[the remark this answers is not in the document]
fmigneault: added by 5cabd1b

docs/services.rst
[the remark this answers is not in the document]
fmigneault: Yes. Thus the warning.
It actually applies to any resource (that allows sub-children) that is located at the lowest *current* level of the tree, regardless of service/resource-type. Adding recursive allow access to resources that could be added at a later time under it.

dbyrns (approved):
Everything's fine with me! Thank you.

fmigneault:
@dbyrns redeployed on ogc-ades

PR #366, opened 2020-11-10
add catalog.* as exception prefix for ServiceTHREDDS

fmigneault:
- adjust patterns to use shell syntax (more natural against dir/file of thedds + avoid ambiguous dot as *any* ) 
- fix edge-case where `thredds/catalog.html`, `thredds/catalog.xml`, etc. (for top-level access), would attempt match against the `prefix` part of other URL and would fail, although equivalent paths `/thredds` and `/thredds/catalog/catalog.html` would correctly work (as expected None/catalog `prefix` is matched.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,10 +9,19 @@ Changes
 
 Features / Changes
 ~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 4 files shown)


fmigneault:
see #367

PR #368, opened 2020-11-16
fix permission set from config

fmigneault:
- fix missing execution of register tests 
- fix handling of  JSON/str explicit permission definition from config

### DIFF
--- a/config/permissions.cfg
+++ b/config/permissions.cfg
@@ -12,6 +12,7 @@
 # -----------
 #   service:              service name to receive the permission (directly on it if no 'resource' mentioned, must exist)
 #   resource (optional):  tree path of the service's resource (ex: /res1/sub-res2/sub-sub-res3)
+#   type (optional):      resource type employed in case of ambiguity between multiple choices (depends on service type)
 #   user and/or group:    user/group for which to apply the permission (create if missing, see below)
 #   permission:           name or object of the permission to be applied (see 'magpie.permissions' for supported values)
 #   action:               one of [create, remove] (default: create)
@@ -29,9 +30,12 @@
 # Default behaviour:
 # ------------------
 #   - create missing resources if supported by the service (and tree automatically resolvable), then apply permissions.
-#   - create missing user/group if required (default user created: (group: anonymous, password: 12345).
-#   - applicable service, user or group is missing, corresponding permissions are ignored and not updated.
-#   - unknown actions are ignored and corresponding permission are not updated, unspecified action resolves to 'create'.
+#   - create missing user/group if required (default user created: (group: anonymous, password: 12345) ).
+#   - if applicable service, user or group is missing, corresponding permissions are ignored and not updated.
+#   - if resolved resource during 'create' action results into an ambiguity (eg: cannot infer which type to apply),
+#     the operation is skipped and the corresponding permission is not applied (use 'type' field to specify it).
+#   - unknown actions are ignored and corresponding permission are not updated.
+#   - unspecified action defaults to 'create'.
 #   - already satisfied permission configurations are left as is
 #     (ie: existing permission to create or non-existing permission to remove are passed)
 #   - environment variables (formatted as `$name` or `${name}`) are expanded if they are matched in the environment.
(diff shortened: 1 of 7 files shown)


dbyrns (approved):

config/permissions.cfg
```diff
@@ -29,9 +30,12 @@
 # Default behaviour:
 # ------------------
 #   - create missing resources if supported by the service (and tree automatically resolvable), then apply permissions.
-#   - create missing user/group if required (default user created: (group: anonymous, password: 12345).
-#   - applicable service, user or group is missing, corresponding permissions are ignored and not updated.
-#   - unknown actions are ignored and corresponding permission are not updated, unspecified action resolves to 'create'.
+#   - create missing user/group if required (default user created: (group: anonymous, password: 12345) ).
```
tlvu, line 33: Curious about this user created with hardcode password, can the user login with this hardcoded password?  Or is the user disabled until a real password is set by the admin?

MatProv (approved):

config/permissions.cfg
```diff
@@ -29,9 +30,12 @@
...
+#   - if applicable service, user or group is missing, corresponding permissions are ignored and not updated.
+#   - if resolved resource during 'create' action results into an ambiguity (eg: cannot infer which type to apply),
```
tlvu, line 35: "if resolved resource during 'create' action results into an ambiguity (eg: cannot infer which type to apply)" So that was the root cause of the failure in PAVICS?

tests/test_register.py
```diff
@@ -271,7 +301,75 @@ def test_register_permissions_existing_group_without_intermediate_entries(self):
...
+        utils.TestSetup.create_TestService(self,
+                                           override_service_name=svc_name,
+                                           override_service_type=svc_type)
+        utils.TestSetup.create_TestGroup(self, override_group_name=self.test_perm_grp_name)
+        session = get_db_session_from_settings(self.app.app.registry.settings)
+
+        res1_name = "test-resource"
+        res2_name = "sub-test-resource"
+        res3_name = "sub-sub-test-resource"
+        res3_perm = "write-deny-recursive"
```
tlvu, line 321: Only test write, not read and browse?  Because if one work all the rest guaranty to work?

tests/test_register.py
```diff
@@ -271,7 +301,75 @@ def test_register_permissions_existing_group_without_intermediate_entries(self):
...
+        # check that all service, resources, group are created correctly
+        services = utils.TestSetup.get_RegisteredServicesList(self)
+        utils.check_val_is_in(svc_name, [s["service_name"] for s in services])
+        groups = utils.TestSetup.get_RegisteredGroupsList(self)
+        utils.check_val_is_in(self.test_perm_grp_name, groups)
+        resp = utils.test_request(self.app, "GET", "/services/{}/resources".format(svc_name))
+        body = utils.check_response_basic_info(resp)
+        svc_res = body[svc_name]["resources"]  # type: JSON
+        utils.check_val_is_in(res1_name, [svc_res[r]["resource_name"] for r in svc_res])
+        res1_id = [svc_res[r]["resource_id"] for r in svc_res if svc_res[r]["resource_name"] == res1_name][0]
```
tlvu, line 347: This block looks a lot like in test_register_permissions_existing_group_create_some_entries, possibility to factor out some code?

tlvu (approved):
LGTM.  I left some comments but since I know nothing about the code, sorry if the comments are not relevant.

config/permissions.cfg
fmigneault: For now the user can login with that password. It is just like any other user creation. 
There is DACCS story about tmp password. This is one use case that is addressed by it.

config/permissions.cfg
fmigneault: the parsing of `read-deny-recursive` was compared against the plain permssion `read` without considering the access/scope modifiers

tests/test_register.py
fmigneault: Yes. This test is more to validate the function that parses configs into resource/permission combinations. 
Once this is done, the requests called are the same as all over the other tests cases that validate the various permission combinations.

Issue #370, opened 2020-11-19
[BUG] Edit/update password of user via UI with smaller length than minimum produces UI 500 error


fmigneault:
**Describe the bug**
See #369 for details. 

**To Reproduce**
Change user password with smaller length than configured. 

**Expected behavior**
Should be handled as insufficient length before submit to API (as done in user creation view).

**Screenshots**
see #369 

**Environment**
 - Version `3.x`

Issue #369, opened 2020-11-19
[Feature] Better error on the web ui when changing password


tlvu:
**Is your feature request related to a problem? Please describe.**
Not useful error on web ui when changing password.
Magpie server `3.2.1`, from PR https://github.com/bird-house/birdhouse-deploy/pull/102

![2020-11-18-220630_727x405_scrot](https://user-images.githubusercontent.com/11966697/99616128-ac9d2d00-29ea-11eb-82b8-ef2deb2b65d7.png)

Corresponding `docker logs magpie`:
```
[2020-11-19 03:02:11,033] ERROR      [ThreadPoolExecutor-0_0][magpie.ui.utils] Unexpected API error under UI operation. [{'param': {'conditions': {'not_in': False}, 'value': 6, 'name': 'password', 'compare': 'range(0, 12)'}, 'code': 400, 'detail': "Invalid 'password' length specified (<12 characters).", 'type': 'application/json', 'path': '/users/admin-catalog', 'url': 'https://lvupavicsdev.ouranos.ca/magpie/users/admin-catalog', 'method': 'PATCH'}]
```

**Describe the solution you'd like**
Should have reported clearly the root cause: "Invalid 'password' length specified (<12 characters).".

fmigneault:
>  Not useful error on web ui when changing password.

I would argue that in this case, the UI is intended to display something that tells in advance the password is invalid instead of attempting to update it with invalid value. So in that regard, the message is correct to display 500 because it wasn't supposed to reach that operation yet. Definitely a bug though.

PR #371, opened 2020-11-20
Fix UI user edit fields error message

fmigneault:
- fix UI error pre-validation of user fields (#369, #370)
- adjust rendered UI error in case of non-HTTP related error
- adjust HTML/CSS mismatches between various forms that were intended to be of 'similar' display format

Example below of displayed contextual error after [password 'Edit', set invalid value, 'Save' button] sequence

image


dbyrns (approved):

tlvu (approved):
The new screenshot with the error LGTM :D

PR #372, opened 2020-11-21
Store custom config JSON in db 

fmigneault:
@tlvu @dbyrns 

## Context 

Regarding comments in https://github.com/bird-house/birdhouse-deploy/pull/102

This PR adds the JSON config from `providers.cfg` / `config.yml` and stores it directly into postgres. 
This allows Twitcher to retrieve the same config later on.

Below samples use this config: 
``` yaml
providers:
  TestThredds:
    url: http://${HOSTNAME}:8000/thredds
    type: thredds
    configuration:
      file_patterns:
        # note: make sure to employ quotes and double escapes to avoid parsing YAML error
        - ".+\\.nc"
      metadata_type:
        prefixes:
          - null  # note: special YAML value evaluated as `no-prefix`, use quotes if literal value is needed
          - "\\w+\\.gif"  # favicon, threddsIcon, etc.
          - "\\w+\\.txt"  # licence
          - "\\w+\\.css"  # tds.css
          - "catalog\\.\\w+"  # note: special case for `THREDDS` top-level directory (root) accessed for `BROWSE`
          - catalog
          - ncml
      data_type:
        prefixes:
          - fileServer
          - dodsC
          - dap4
```

As shown on the right side, I can successfully access the resources (with anonymous recursive applied on TestThredds)

image

Now, there is still an issue with THREDDS as illustrated above. PAVICS/DACCS stack registers THREDDS service with the prefix `/thredds` within the URL in Magpie, as done in my example. All references of THREDDS UI assume to be at the root (without the prefix) and therefore point to wrong location 
(aka `<>/twitcher/ows/proxy/TestThredds/thredds/favicon.ico` becomes `<real THREDDS>/thredds/thredds/favicon.ico`). 

image

The clean thing to do is therefore to register *without* the `/thredds` prefix in Magpie, but then every request via Twitcher must be as follows: `<>/twitcher/ows/proxy/TestThredds/thredds/...`. 
In the case of PAVICS/DACCS, this looks a bit odd because the service-name is also `thredds`, making request formatted like such: 
`<>/twitcher/ows/proxy/thredds/thredds/...`

I can correctly access the icons using this method, but it seems Thredd's UI insists on using the wrong root
image

## Question

What do we do? 

1. We modify the URL of DACCS, but this will probably impact a lot of references that already assume only one `/thredds`.
2. "Too bad" for icons?
3. Rebase the corresponding icons/css within real Thredds to they can be found under `/thredds/thredds`? 
4. other idea?

From Magpie's point of view, I cannot do much more. It is the service that sends the hardcoded `/thredds/...` UI requests.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -22,6 +22,9 @@ Features / Changes
   ``magpie.constants.get_constant``, avoiding inconsistent resolution of setting value versus environment variable
   wherever the settings container was not passed down everywhere over deeply nested function calls.
 * Handle `Twitcher`, `PostgreSQL` and `Phoenix` setting prefix conversion from corresponding environment variable names.
+* Store custom configuration of ``Service`` into database for same definition retrieval between `Magpie` and `Twitcher`
+  without need to provide the same configuration file to both on startup.
+* Display custom ``Service`` configuration as JSON/YAML on its corresponding UI edit page when applicable.
 
 Bug Fixes
 ~~~~~~~~~~~~~~~~~~~~~
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -16,6 +16,12 @@ Features / Changes
 * Align HTML format and structure of all edit forms portions of `Users`, `Groups` and `Services` UI pages to simplify
   and unify their rendering.
 * Add inline UI error messages to `User` edition fields.
+* Improve resolution of `Twitcher` URL using ``TWITCHER_HOST`` explicitly provided  setting (or environment variable)
+  before falling back to default ``HOSTNAME`` value.
+* Employ `Pyramid`'s local thread registry to resolve application settings if not explicitly provided to
+  ``magpie.constants.get_constant``, avoiding inconsistent resolution of setting value versus environment variable
+  wherever the settings container was not passed down everywhere over deeply nested function calls.
+* Handle `Twitcher`, `PostgreSQL` and `Phoenix` setting prefix conversion from corresponding environment variable names.
 
 Bug Fixes
 ~~~~~~~~~~~~~~~~~~~~~
--- a/Makefile
+++ b/Makefile
@@ -208,7 +208,7 @@ database-history: conda-env _alembic    ## obtain database revision history
 .PHONY: database-revision
 database-revision: conda-env _alembic   ## create a new database revision
 	@[ "${DOC}" ] || ( echo ">> 'DOC' is not set. Provide a description."; exit 1 )
-	@bash -c '$(CONDA_CMD) alembic -c "$(APP_INI)" revision -m $(DOC)'
+	@bash -c '$(CONDA_CMD) alembic -c "$(APP_INI)" revision -m "$(DOC)"'
 
 .PHONY: database-version
 database-version: conda-env _alembic 	## retrieve current database revision ID
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -613,7 +613,8 @@ Following settings define parameters required by `Twitcher`_ (OWS Security Proxy
   .. versionadded:: 2.0
 
   Specifies the explicit hostname to employ in combination with ``TWITCHER_PROTECTED_PATH`` to form the complete base
-  service protected URL. Ignored if ``TWITCHER_PROTECTED_URL`` was provided directly.
+  service protected URL. Ignored if ``TWITCHER_PROTECTED_URL`` was provided directly. If not provided, hostname
+  resolution falls back to using ``HOSTNAME`` environment variable.
 
   .. note::
     The resulting URL will take the form ``https://{TWITCHER_HOST}[/twitcher]{TWITCHER_PROTECTED_PATH}`` to imitate
--- a/docs/services.rst
+++ b/docs/services.rst
@@ -217,31 +217,41 @@ by the below configuration.
         type: thredds
 
         # customizable request parsing methodology (specific to `thredds` type)
-        file_patterns:
-          # note: make sure to employ quotes and double escapes to avoid parsing YAML error
-          - ".*\\.nc"
-        metadata_type:
-          prefixes:
-            - null  # note: special YAML value evaluated as `no-prefix`, use quotes if literal value is needed
-            - "catalog\\.\\w+"  # note: special case for `THREDDS` top-level directory (root) accessed for `BROWSE`
-            - catalog
-            - ncml
-            - uddc
-            - iso
-        data_type:
-          prefixes:
-            - fileServer
-            - dodsC
-            - dap4
-            - wcs
-            - wms
+        configuration:
+          # path prefix to skip (strip) before processing the rest of the path in case the
+          # registered service URL in Magpie does not have the same root as proxied by Twitcher public URL
+          skip_prefix: thredds
+          # define which pattern matches that will map different path variations into same file resources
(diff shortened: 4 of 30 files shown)


tlvu:
@fmigneault Sorry I am lost as to why there would be a duplicate "thredds" in the URL.

Currently on Boreas production, this is the working icon link https://pavics.ouranos.ca/twitcher/ows/proxy/thredds/threddsIcon.gif.

As you can see, the part `/twitcher/ows/proxy/thredds/` is managed by Twitcher/Magpie, then `threddsIcon.gif` immediately follows so there is no duplicate "thredds" there.

Another basic question is why does Twitcher needs to know Magpie config?  I assume Twitcher give Magpie a link and Magpie answer whether the link can be accessible for the specified user.  Magpie is making the call, Twitcher just execute.

I see the relationship between Twitcher and Magpie as a login page using an external authentication provider.  The external authentication provider make the call that the user and password is valid and the login page let the user in, no need to know any configs about how the external auth provider works.

dbyrns:
Like @tlvu said, @fmigneault I don't think we need a duplicate 'thredds' in the URI. As for thredds insisting with the /thredds root, I think this is a config on the thredds side. For your scenario, I guess that we could make it works if thredds was configured with the root being "TestThredds". Then we should be able to retrieve the icon at <>/twitcher/ows/proxy/TestThredds/favicon.ico

@tlvu, when @fmigneault says that twitcher need to know something about magpie, what he really says is that the MagpieAdapter of twitcher, which implement the auth interface, needs to access the configuration to properly handle a request. @fmigneault correct me if I'm wrong, but the MagpieAdapter deals with requests directly from twitcher processes and doesn't call Magpie API to resolve the security. So, twitcher instance, because of the MagpieAdapter, needs to access Magpie's configuration.

We have discussed that aspect in the past, and I'm not sure if we still have plans to move everything on the Magpie side and use API calls to resolve requests. The answer is maybe, we won't because of performance issues.

fmigneault:
> Sorry I am lost as to why there would be a duplicate "thredds" in the URL.


The first `/thredds` is the reference name of the service (which just happends to be `thredds` but could be anything), while the second `/thredds` is the prefix of the real thredds service (the root is its tomcat process). 

> Another basic question is why does Twitcher needs to know Magpie config? I assume Twitcher give Magpie a link and Magpie answer whether the link can be accessible for the specified user. Magpie is making the call, Twitcher just execute.

Twitcher doesn't ask Magpie via API whether it should grant access or not, but rather uses the MagpieAdapter with direct database access. So if the loaded *custom* config in Magpie for a given Thredds service doesn't match what is loaded by Twitcher, the authorization resolution is not done the same way. To resolve this issue, we would need to load the exact same file by both Twitcher and Magpie. Instead, this PR stores the config in the DB so that Twitcher simply uses the service config that was previously loaded by Magpie without loading the config file.

dbyrns:
> The first /thredds is the reference name of the service (which just happends to be thredds but could be anything), while the second /thredds is the prefix of the real thredds service (the root is its tomcat process).

I think both magpie and tomcat are configured using the same name exactly to avoid the duplication. So this is not a coincidence at all that the service is named "thredds".

fmigneault:
I'll be honest, I don't know what kind of trickery PAVICS stack is doing, but I can't replicate. 
Even using the same `thredds` name to avoid duplication, I can access everything correctly, but icons just refuse to be displayed/found via the catalog interface. 


image

tlvu:
> I'll be honest, I don't know what kind of trickery PAVICS stack is doing, but I can't replicate.

What trickery, what observed vs actual behaviour are confusing you?

For me I wonder why the behavior changed since the previous Magpie?  Was that behaviour change planned/needed?  Because on hirondelle (https://hirondelle.crim.ca/thredds/) all the icons are also showing  up fine.  No hidden trick here, it's fully reproducible as you can see.

Putting the new Magpie into the same stack with the exact same config, the icons disappear.

If the fix to the disappearing icons is to have double the `/thredds` then why the previous version of Magpie do not need it?

tlvu:
> Twitcher doesn't ask Magpie via API whether it should grant access or not, but rather uses the MagpieAdapter with direct database access. So if the loaded _custom_ config in Magpie for a given Thredds service doesn't match what is loaded by Twitcher, the authorization resolution is not done the same way. To resolve this issue, we would need to load the exact same file by both Twitcher and Magpie. Instead, this PR stores the config in the DB so that Twitcher simply uses the service config that was previously loaded by Magpie without loading the config file.

I am okay with serializing the config to the DB for sharing, if that's how it's designed to work.

Just make sure the config in the DB is always in sync with the real config file and Twitcher always reload the up-to-date config from DB instead of only reload on restart.

dbyrns:
The trickery is here : https://github.com/Ouranosinc/PAVICS/commit/2959ccdf427be921cea33d007cfed983a35298cd
In fact tomcat is configured with the /twitcher/ows/proxy/thredds as a root so both twitcher and thredds share the exact same URI both using different ports.
twitcher listen here : <>:443/twitcher/ows/proxy/thredds
thredds listen here : <>:8083/twitcher/ows/proxy/thredds

fmigneault:
@tlvu @dbyrns 
Icons are displayed using `daccs-iac` and https://github.com/bird-house/birdhouse-deploy/tree/magpie-thredds
built on top of https://github.com/bird-house/birdhouse-deploy/pull/102 and using this `3.3.0-dev` version (c5d2904)

image

fmigneault:
preview of config in service page: 

image

tlvu:
> preview of config in service page:

Just a curious question, why serialize to json and not yaml?  You losing all the comments in the original yaml.

tlvu:
> Icons are displayed using `daccs-iac` and https://github.com/bird-house/birdhouse-deploy/tree/magpie-thredds
> built on top of [bird-house/birdhouse-deploy#102](https://github.com/bird-house/birdhouse-deploy/pull/102) and using this `3.3.0-dev` version ([c5d2904](https://github.com/Ouranosinc/Magpie/commit/c5d2904efa30b48b05b5d3094d0bef04babd2bc7))
> 

Just want to say I am very happy you are using the daccs-iac to test.  Testing in the exact same config as prod is the most reliable way to test.  That's why I previously did the Vagrant box to replicate prod locally.

DavidB finally explained to me what you called 'trickery PAVICS stack is doing'.  Given the daccs-iac, we know for sure all the config are up-to-date.

So from magpie/providers.cfg here https://github.com/bird-house/birdhouse-deploy/blob/cc4ca20665e5eeb3d3ef51e1588521f4b42bef4d/birdhouse/config/magpie/providers.cfg.template#L58-L59 we can deduce the same info as DavidB:
```
twitcher listen here : <>:443/twitcher/ows/proxy/thredds
thredds listen here : <>:8083/twitcher/ows/proxy/thredds
```

dbyrns (approved):
Gicen the daccs-iac and the screenshots, I'm happy to approve this PR.
Also, this is nice to see the service config directly in the UI.
Thanks

fmigneault:
> Just a curious question, why serialize to json and not yaml? You losing all the comments in the original yaml.

Even when loading the config files with YAML, the comments are dropped when converted to python dicts/lists.

I prefer regardless to do it this way to ensure minimal validation of the format. 
I don't want to dump to db literal comments that could introduce invalid characters/format without making sure they are valid for the service.

PR #373, opened 2020-11-30
group priorities

fmigneault:
working on ... 
- https://github.com/bird-house/birdhouse-deploy/pull/107
- https://github.com/Ouranosinc/PAVICS-e2e-workflow-tests/pull/55

... I encounter yet another edge case

When we define the following structure: 
```
service: Thredds               group: anonymous; perms: browse, read, write; scope: recursive; access: ALLOW
  resource: birdhouse
    resource: secure           group: anonymous; perms: browse, read, write; scope: recursive; access: DENY
                               group: secured; perms: browse, read, write; scope: recursive; access: ALLOW
```

The behaviour of Magpie is to deny a user member of `secured` because the first permission found is for `anonymous` group, which denies access. The following `secured` ALLOW is simply ignored because DENY overrides (always prioritized).

So this PR adds the group `priority` field. 
The idea is simple, the higher that value is, the more is overrides previous decisions **on the same level**.
By default, `anonymous` group is -1, and every other group is >= 0. 

## TODO
- [x] ~listing of groups ordered by priority~ **not important as it is not really natural to guess that order just from names**
- [x] ~API POST/PUT to update group priority~ **wont implement**
- [x] ~UI field to display/update priority~ **not needed as above not implemented**
- [x] ~validation of values (>=0, int)~ **not needed, no input**
- [ ] complete doc example of permission resolution (section `Permissions -> Resolution of Permissions`)


fmigneault:
@dbyrns I would likes your thoughts on this before I go ahead for complete implementation of the todos.

dbyrns:
As discussed this morning, I think that the PR is good enough as is (anonymous -1, other 0). 
It makes sense that denying anonymous doesn't mean denying everybody, but only those who are not part of the allowed groups. For any other groups, it could lead to a complex handling and a rabbit hole we don't want to explore. So I would not show the group priority, don't let the priority be set by the API and don't update the UI. As for the validation >= 0, I guess that it can be avoid too considering the previous statements.

Thank you for this analysis.

fmigneault:
@dbyrns I would like your usual feedback. 
@MatProv / @ChaamC / @tlvu you're also welcome to comment if you see anything up for improvement

PR is "essentially ready", I only got the documentation example part remaining that I can cook up while this review is ongoing. 
Got a test that builds various combinations and returns the expected `reason` value (sadly running only locally now :/)
*generic* group seems correctly prioritized over anonymous (as in public vs secured THREDDS directory case)

magpie/permissions.py
```diff
@@ -128,13 +154,40 @@ def __ne__(self, other):
...
+        We then obtain two **crucial** ordering results:
+            1. We can easily pick the last sorted item with highest resolution priority to find the final result of
+               corresponding permissions.
+               (note: final result for same user or group, their direct/inherited resolution is not considered here).
+            2. Picking the first element with lowest priority also displays the permission that impacts the widest
+               range of resources. For instance in Magpie UI, indicating that a permission as :attr:`Scope.RECURSIVE`
+               is more verbose as it tell other resources under it are also receive the specified :class:`Access`
+               modifier rather than only the punctual resource.
+
+        .. warning::
```
dbyrns, line 189: I wonder, if the warning should not be duplicated at the top too given how bad a modification could be.

magpie/permissions.py
```diff
@@ -204,6 +260,113 @@ def ace(self, user_or_group):
...
+
+    @property
+    def group_priority(self):
+        # type: () -> Optional[Union[int, Type[math.inf]]]
+        """
+        Priority accessor in case of group inherited permission resolved by :class:`PermissionTuple`.
+        """
+        if self._tuple is not None and self.type == PermissionType.INHERITED:
+            if self._tuple.group.group_name == get_constant("MAGPIE_ANONYMOUS_GROUP"):
+                return -1
```
dbyrns, line 356: This special handling answers my question on the database upgrade. But if anonymous and administrator groups were set in the db you would not have to handle them here. Unless those groups are not even in the db?

dbyrns (commented):
I only got a couple of questions/comments. Overall this PR is good!

docs/permissions.rst
```diff
@@ -235,26 +238,289 @@ to execute `Magpie` application. Effectively, when the active session correspond
...
+normal step conditions and prioritizing :attr:`Access.DENY`.
+
+When resolving only :term:`Inherited Permissions`, the procedure stops here and provides the applicable result if any
+was found, with the corresponding ``reason``. An empty set of :term:`Permission` is returned if none could be found.
+
+When instead resolving :term:`Effective Permissions`, the above process continues by rewinding the parent
+:term:`Resource` hierarchy until the first :term:`Permission` is found. Only on the first iteration (when the targeted
+:term:`Resource` is the same as the one looked for potential :term:`Inherited Permissions`) does :attr:`Scope.MATCH`
+take effect. Only :attr:`Scope.RECURSIVE` are considered afterwards. When a :term:`Permission` is found, the process
+immediately completes if :attr:`Access.DENY` results from the previous resolution steps. Otherwise, the process still
```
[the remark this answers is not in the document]
fmigneault: If the found permission is for User, yes allow/deny is returned directly because that the highest possible priority, you are right.
This actually made me think of one still case not handled... 

If we find a Group Allow/Deny, we must continue rewinding regardless in case there is a User Deny/Allow-Recursive higher up that reverts it. 
I could confirm that this doesn't work currently.

docs/permissions.rst
```diff
@@ -235,26 +238,289 @@ to execute `Magpie` application. Effectively, when the active session correspond
...
+        "permission_names": [
+            "[permission-name]",
+            "[name-access-scope]"
+        ],
+        "permissions": [
+            {
+                "name": "permission-name",
+                "access": "allow|deny",
+                "scope": "match|recursive",
+                "type": "access|allowed|applied|direct|inherited|effective|owned",
```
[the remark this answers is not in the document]
fmigneault: Yes, the field is mentionned somewhere on top of where `reason` is described. The full list of each type is completely at the top of the same file.

magpie/alembic/versions/2020-11-30_9d977cce1ef2_group_permission_priority.py
```diff
@@ -0,0 +1,45 @@
...
+    session = Session(bind=op.get_bind())
+
+    # two following lines avoids double "DELETE" erroneous call when deleting group due to incorrect checks
+    # https://stackoverflow.com/questions/28824401
+    context.connection.engine.dialect.supports_sane_rowcount = False
+    context.connection.engine.dialect.supports_sane_multi_rowcount = False
+
+    if isinstance(context.connection.engine.dialect, PGDialect):
+        op.add_column("groups", sa.Column("priority", sa.Integer(), nullable=False, default=0, server_default="0"))
+        stmt = groups.update().where(groups.c.priority is None).values({"priority": 0})
```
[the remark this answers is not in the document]
fmigneault: I decided to go with code override since technically we can boot with different `MAGPIE_ANONYMOUS_GROUP` value and have the literal value change name. So whatever is stored in the db doesn't matter that much anymore. I'm having also some trouble defining an `inf` for admins. 
Maybe I should remove the migration script altogether and remove the priority db field to use explicit if against the name and return -1/0/1 ?

magpie/permissions.py
fmigneault: I'll move it toward the top so it is the second thing listed after minimal description. Harder to miss. I'll also describe it a bit better.

magpie/permissions.py
fmigneault: They are in the db and would indeed cause misbehavior. This is why I just didn't write anything at all in the db.
I feel even more like reverting the migration. It will be just a plain property against the instance group-name.

docs/permissions.rst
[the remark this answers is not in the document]
dbyrns: Good point. But given that the current code doesn't account for user permissions upward, I'm wondering why do you rewind right now? Am I missing something else?

magpie/alembic/versions/2020-11-30_9d977cce1ef2_group_permission_priority.py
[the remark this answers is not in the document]
dbyrns: Given that we agreed that we want to avoid the standard groups priority hell, I guess that we could indeed keep that door fully close and don't add this field.

docs/permissions.rst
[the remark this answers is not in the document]
fmigneault: doc fixed with latest resolution method

docs/permissions.rst
```diff
@@ -471,31 +497,113 @@ access privileges than its unauthenticated counterpart in case of contradictory
...
-:term:`Resource` is the same as the one looked for potential :term:`Inherited Permissions`) does :attr:`Scope.MATCH`
-take effect. Only :attr:`Scope.RECURSIVE` are considered afterwards. When a :term:`Permission` is found, the process
-immediately completes if :attr:`Access.DENY` results from the previous resolution steps. Otherwise, the process still
-continues until reaching the top-most :term:`Service` to validate :attr:`Access.ALLOW` over the entire scope the
-:term:`Permission` was found. If still no :term:`Permission` is defined after complete hierarchy processing, the result
-defaults to :attr:`Access.DENY`, and indicated by ``"no-permission"`` reason.
-
-.. todo: pseudo-code to represent the 'effective permission' portion of the resolution
-
+When instead resolving :term:`Effective Permissions`, there is an additional pre-step to (1) that verifies if
```
dbyrns, line 507: I love the new documentation. I'm just wondering why you didn't respect the chronological order (admin as step 1 and rewinding as a step, so that we get 4 steps : 

1. admin
2. direct
3. inherited
4. rewinding

dbyrns (approved):
Good job, we have a winner!
Feel free to update the documentation as I suggested or not, it's up to you.

docs/permissions.rst
fmigneault: I can adjust it with these steps to make it clearer.

dbyrns: And this is the same for the rewinding step, I did not catch this particularity. Maybe you could do something like that : 
1. [effective only] admin
2. [all type] direct
3. [inherited and effective] inherited
4. [effective only] rewinding

dbyrns:
Good! Ready to merge.

PR #374, opened 2020-12-02
User self delete

ChaamC:
This PR adds the option for a user to delete his own account via the ui. It includes multiple test cases to delete a user account.
Also, some doc has been added to help on configuring magpie to run the tests.

image

image


dbyrns:
LGTM. Add a mention that all data linked to this account will also be lost.

tests/interfaces.py
```diff
@@ -4442,6 +4639,40 @@ def test_UserAccount_UpdateDetails_password(self):
...
         utils.check_response_basic_info(resp, 401, expected_method="POST")
 
+    @runner.MAGPIE_TEST_USERS
+    @runner.MAGPIE_TEST_LOGGED
+    @runner.MAGPIE_TEST_FUNCTIONAL
+    def test_UserAccount_DeleteSelf(self):
+        """
+        Logged user can delete his own account on account page.
+        """
+        utils.warn_version(self, "user account page", "3.3.1", skip=True)
```
fmigneault, line 4649: add 'delete' to the string to print the appropriate reason

magpie/ui/user/templates/edit_current_user.mako
```diff
@@ -30,23 +30,46 @@
...
-            <input type="submit" value="Delete Account" name="delete" class="button delete">
+            <input value="Delete Account" name="delete"
+				%if user_name in MAGPIE_FIXED_USERS:
+					class="button delete disabled" type="button" disabled
+				%else:
+				    type="button"
+                    class="button delete"
+                    onclick="document.getElementById('EditService_UserSelfDeleteAlert').style.display = 'block'"
+				%endif
+			>
```
fmigneault, line 71: please align the contents

magpie/ui/user/templates/edit_current_user.mako
```diff
@@ -30,23 +30,46 @@
 
 <h3>User Information</h3>
 
+<div class="alert alert-danger" id="EditService_UserSelfDeleteAlert">
+    <h3 class="alert-title-danger">Danger!</h3>
+    <div class="alert-info">
```
fmigneault, line 35: I think one of the class needs to be adjusted here. 
The demo image shows a blue block area that should be red.

docs/usage.rst
```diff
@@ -97,3 +97,27 @@ Please refer to this section for more details.
...
+
+To execute the tests on your local machine, a PostgreSQL instance must be running.
+This PostgreSQL instance can be easily initialized by using the docker-compose.yml.example file with the default values.
+Make sure the values defined in postgres.env are also corresponding to your testing setup.
+
+There are both local and remote tests. The PostgreSQL is sufficient for the local tests but the remote tests
+also require a running instance of Magpie.
+
+A basic Magpie instance can also be initialized by using the docker-compose.yml.example file.
+Note that the environment variables MAGPIE_TEST_REMOTE_SERVER_URL and HOST_FQDN
```
fmigneault, line 112: use double `` around the variables.

docs/usage.rst
```diff
@@ -97,3 +97,27 @@ Please refer to this section for more details.
 .. _installation: installation.rst
 .. _permissions: permissions.rst
 .. _utilities: utilities.rst
+
+Testing
+~~~~~~~
+
+To execute the tests on your local machine, a PostgreSQL instance must be running.
+This PostgreSQL instance can be easily initialized by using the docker-compose.yml.example file with the default values.
```
fmigneault, line 105: use references for `PostgreSQL` and `docker-compose.yml.example`
you can check in configuration.rst, they are already defined in there.

env/postgres.env.example
```diff
@@ -1,5 +1,5 @@
-MAGPIE_POSTGRES_USERNAME=postgres
+MAGPIE_POSTGRES_USERNAME=magpie
 MAGPIE_POSTGRES_PASSWORD=qwerty
-MAGPIE_POSTGRES_HOST=postgres
+MAGPIE_POSTGRES_HOST=localhost
```
fmigneault, line 3: I would had a note because this is probably one of the most confusing thing usually. 
If running docker magpie and using a local postgres, localhost is right. 
But inside a docker-compose stack, the name of the linked postgres image must be used to find the network.

tests/interfaces.py
```diff
@@ -832,6 +884,72 @@ def test_UpdateUsers_username_Forbidden_UpdateOthers(self):
         resp = utils.test_request(self, "GET", path, cookies=self.cookies, headers=self.json_headers)
         utils.check_response_basic_info(resp)
 
+    @runner.MAGPIE_TEST_USERS
+    def test_DeleteUser_DeleteSelf(self):
+        """
+        Logged user is allowed to delete his own account.
+        """
+        # login as test user and delete its own account
```
fmigneault, line 892: add skip safeguard, 
other tests are ok because 403 remains 403.
but this one is a change of behavior

CHANGES.rst
```diff
@@ -9,6 +9,17 @@ Changes
 
 * Nothing yet.
 
+`3.3.1 <https://github.com/Ouranosinc/Magpie/tree/3.3.1>`_ (2020-12-02)
+------------------------------------------------------------------------------------
```
fmigneault, line 13: revert only the header part to keep 'latest' and remove the tmp version
we will do an actual bump in master to generate the tags/dockers

fmigneault (commented):
a few linting patches but overall LGTM

magpie/ui/user/templates/edit_current_user.mako
ChaamC: Changed it to the red background, pics here :

image

fmigneault: Can you check also for the class that causes the "newline" of the text. 
The (!) should be on the same line followed by "Delete your account".

I think the problem is caused by same css class being reuse in 2 locations that require different configs.

tests/interfaces.py
ChaamC: Should I also add a skip safeguard to the corresponding test on the admin scope? Or was this an existing behavior for the admin level?

https://github.com/Ouranosinc/Magpie/blob/859340a9771b1995212b8243e12dc8444212a2f2/tests/interfaces.py#L3088-L3093

CHANGES.rst
ChaamC: So, that would mean changing the highlighted line to this :
?
I wasn't sure about the good practice here with version bumping. I was wondering if I should have done it after the PR was approved. The problem was I wanted to reference the version in the test file interfaces.py, for the skip safeguards, so I bumped the version to make it clear I was referencing the current version in the skip safeguards.

magpie/ui/user/templates/edit_current_user.mako
ChaamC: How does this look? (also added an extra space between the (!) and the text)

image

tests/interfaces.py
fmigneault: The admins were allowed to self delete before as they can do anything (except remove special anonymous/admin).
So the result should be the same.

CHANGES.rst
fmigneault: Yes, I understand what you mean. I do the same kind of hack locally for testing new versions. 
The only thing is that since there isn't an actual tag and version, I don't want them to be displayed in the doc pushed to readthedocs when it gets merged. 

I haven't found a clean way to solve this yet.

Also, I just noticed, your compared should use `3.4.0` instead of `3.3.1`.

fmigneault: try with this patch: 
https://github.com/Ouranosinc/Magpie/tree/test-version

fmigneault (commented):
only the version left to adjust, see relative comment

tests/interfaces.py
ChaamC: Ok! I won't put a safeguard for the admin case then.

PR #375, opened 2020-12-04
MAGPIE_TEST_VERSION envvar

fmigneault:
Add test-version option to ignore safeguards during development of next version.

This is useful when a new test that brings a new feature must define `>= "future-version"` that is not yet deployed.
The tests are not run because the condition is not yet met, which always needed a fake "bump-version" to pass, and then revert for real bump/tag.

Within this, one can set `MAGPIE_TEST_VERSION = "latest"` prior to running tests and all compares to `>= "future-version"` will return `True`.


tests/utils.py
```diff
@@ -178,6 +178,35 @@ def __new__(cls, name, description=None):
         return make_run_option_decorator(RunOption(name, description=description))
 
 
+class TestVersion(LooseVersion):
+    """
+    Special version supporting ``latest`` keyword to ignore safeguard check of :func:`warn_version` during development.
+
+    .. seealso::
+        Environment variable ``MAGPIE_TEST_VERSION`` should be set with the desired version or ``latest`` to evaluate
+        even new features above the last tagged version.
```
ChaamC, line 187: Maybe the MAGPIE_TEST_VERSION could also be documented elsewhere (with an explanation similar to this comment), or added to the docker-compose.yml.example and/or magpie.env.example?

ChaamC (commented):
Apart from the single comment I added, the rest seems good!

tests/utils.py
fmigneault: Yes. I was waiting for you 'tests' section update to add it.

env/magpie.env.example
```diff
@@ -21,10 +23,19 @@ PHOENIX_PORT=8443
 PHOENIX_PUSH=false
 TWITCHER_PROTECTED_PATH=/ows/proxy
 
-# Magpie Tests
+## === Magpie Tests ===
+# below version defines the override value for test
+# when using tes safeguards for different versions, new tests that require 'future' tags yet to be defined can
```
ChaamC, line 28: Small typo : missing 't' from test in "when using tes"

ChaamC (commented):
Good improvements to the test documentation! :)

LGTM

env/magpie.env.example
fmigneault: nice catch ;)

Issue #161, opened 2019-03-27
User creation with dummy password and login update


fmigneault:
Allow creation of a new user with empty password.
In this case: 
- creates a random temporary password and sets a flag `temporary=True`.
- sends a notification email to `email` from the form with `username` and *temporary* `password`.

On first login, if `temporary=True`:
- redirects to an update password page.
- validates that new password != temporary 
- sets the updated encrypted password if valid and `temporary=False`

dbyrns:
Required by the GIN frontend : https://www.crim.ca/jira/browse/GEOIM-104

PR #379, opened 2021-01-05
tmp-url

fmigneault:
* adds tmp URL endpoint to receive token and attempt resolution of the appropriate `TokenOperation`.
* adds a table of `TemporaryToken` for pending operations mapped to tmp URL endpoint to "complete" it
* adds basic setup of "terms and conditions" validation via tmp URL (relates to [DAC-119](https://www.crim.ca/jira/browse/DAC-119))
  still requires: 
  - actual T&C field in group and corresponding API & UI group create updates
  - email with rendered T&C and embedded accept tmp-URL
* adds basic setup of "password reset" operation
  still requires:
  - actual handling to complete reset procedure (see [DAC-117](https://www.crim.ca/jira/browse/DAC-117)) 
  - sending emails for various use cases and handling responses from endpoints formatted within these emails

* includes also minor fixes of path parameter generation in OpenAPI schemas

### DIFF
--- a/magpie/alembic/versions/2021-01-04_954a9d7fe740_temporary_url_table.py
+++ b/magpie/alembic/versions/2021-01-04_954a9d7fe740_temporary_url_table.py
@@ -1,0 +1,36 @@
+"""
+Temporary URL table
+
+Revision ID: 954a9d7fe740
+Revises: cfe64807c143
+Create Date: 2021-01-04 12:56:22.298527
+"""
+
+import datetime
+import sqlalchemy as sa
+import uuid
+from alembic import op
+from sqlalchemy.dialects.postgresql import UUID
+
+
+# revision identifiers, used by Alembic.
+revision = "954a9d7fe740"
+down_revision = "cfe64807c143"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.create_table("tmp_tokens",
+                    sa.Column("token", UUID(as_uuid=True), primary_key=True, default=uuid.uuid4, unique=True),
+                    sa.Column("operation", sa.Unicode(32), nullable=False),
+                    sa.Column("user_id", sa.Integer,
+                              sa.ForeignKey("users.id", onupdate="CASCADE", ondelete="CASCADE"), nullable=True),
+                    sa.Column("group_id", sa.Integer(),
+                              sa.ForeignKey("groups.id", onupdate="CASCADE", ondelete="CASCADE"), nullable=True),
+                    sa.Column("created", sa.DateTime, default=datetime.datetime.utcnow)
+                    )
+
+
+def downgrade():
+    op.drop_table("tmp_tokens")
--- a/magpie/alembic/versions/2021-01-04_954a9d7fe740_temporary_url_table.py
+++ b/magpie/alembic/versions/2021-01-04_954a9d7fe740_temporary_url_table.py
@@ -1,5 +1,5 @@
 """
-Temporary URL table
+Temporary URL table.
 
 Revision ID: 954a9d7fe740
 Revises: cfe64807c143
--- a/magpie/api/management/register/register_views.py
+++ b/magpie/api/management/register/register_views.py
@@ -100,12 +100,12 @@ def handle_temporary_url(request):
     """
     Handles the operation according to the provided temporary URL token.
     """
-    token = ar.get_value_matchdict_checked(request, key="token", pattern=ax.UUID_REGEX)
-    token = token.split(":")[-1]  # remove optional prefix if any
+    str_token = ar.get_value_matchdict_checked(request, key="token", pattern=ax.UUID_REGEX)
+    str_token = str_token.split(":")[-1]  # remove optional prefix if any (e.g.: 'urn:uuid:')
     db_session = request.db
-    tmp_url = models.TemporaryToken.by_token(token, db_session=db_session)
-    ax.verify_param(tmp_url, not_none=True,
-                    http_error=HTTPNotFound, content={"token": str(token)},
+    tmp_token = models.TemporaryToken.by_token(str_token, db_session=db_session)
+    ax.verify_param(tmp_token, not_none=True,
+                    http_error=HTTPNotFound, content={"token": str(str_token)},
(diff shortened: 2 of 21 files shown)


magpie/api/management/register/register_views.py
```diff
@@ -82,3 +91,21 @@ def leave_discoverable_group_view(request):
...
+                       response_schemas=s.TemporaryURL_GET_responses)
+@view_config(route_name=s.TemporaryUrlAPI.name, request_method="GET", permission=NO_PERMISSION_REQUIRED)
+def handle_temporary_url(request):
+    """
+    Handles the operation according to the provided temporary URL token.
+    """
+    token = ar.get_value_matchdict_checked(request, key="token", pattern=ax.UUID_REGEX)
+    token = token.split(":")[-1]  # remove optional prefix if any
+    db_session = request.db
+    tmp_url = models.TemporaryToken.by_token(token, db_session=db_session)
```
dbyrns, line 106: by_token and handle_temporary_token don't seem to deal with url but rather with an actual TemporaryToken. Variable name is misleading

magpie/api/management/register/register_views.py
```diff
@@ -82,3 +91,21 @@ def leave_discoverable_group_view(request):
     group = ru.get_discoverable_group_by_name(group.group_name, db_session=request.db)
     uu.delete_user_group(user, group, request.db)
     return ax.valid_http(http_success=HTTPOk, detail=s.RegisterGroup_DELETE_OkResponseSchema.description)
+
+
+@s.TemporaryUrlAPI.get(schema=s.TemporaryURL_GET_RequestSchema, tags=[s.RegisterTag],
+                       response_schemas=s.TemporaryURL_GET_responses)
+@view_config(route_name=s.TemporaryUrlAPI.name, request_method="GET", permission=NO_PERMISSION_REQUIRED)
```
dbyrns, line 98: Should we required the user to be logged in to access the temporary endpoint?
What if someone claim that it never accepts the T&C or that someone resets his password for him?
Could the temporary endpoint redirects to login screen (if no current session) before performing the job?

dbyrns (changes requested):
Nice draft!
Should we also add the counterpart of handle_temporary_url which would be something like create_temporary_url ? I guess that it would be better to have the creation process in one place rather than scattered for each use case.

ChaamC (approved):
LGTM, nothing to add!

Will tests be added later, with the tasks for terms & conditions, and password reset?

magpie/alembic/versions/2021-01-04_954a9d7fe740_temporary_url_table.py
```diff
@@ -0,0 +1,36 @@
...
+depends_on = None
+
+
+def upgrade():
+    op.create_table("tmp_tokens",
+                    sa.Column("token", UUID(as_uuid=True), primary_key=True, default=uuid.uuid4, unique=True),
+                    sa.Column("operation", sa.Unicode(32), nullable=False),
+                    sa.Column("user_id", sa.Integer,
+                              sa.ForeignKey("users.id", onupdate="CASCADE", ondelete="CASCADE"), nullable=True),
+                    sa.Column("group_id", sa.Integer(),
```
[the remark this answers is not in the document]
fmigneault: `group_id` and `user_id` are used for the accept T&C `operation`
`user_id` is also used by itself (`group_id = None`) for reset password `operation`

The fields are there is a self-serve fashion. This is why they are allowed to be null.
Whichever information is needed by the `operation` can be defined there, and it up to that `operation` to define expected values properly.
I think the only other "possible" column could be `resource_id`, in case some new `operation` was specific to a resource, but I don't know what that case could be for the moment. 

Since each implementation of `operation` to generate the token are specific to their use case, I think that most of the logic to create the db entry will be in the specific function that needs it.

magpie/api/management/register/register_views.py
fmigneault: Ah, yes, remnant of previous code. Will adjust.

magpie/api/management/register/register_views.py
fmigneault: For T&C, the stored `user_id` should be validated against the logged-user so only he can accept the terms.  
So yes definitely requires to be logged in this case. Redirect to login page as need could be a plus if he wasn't logged in.

For reset-password, the whole point is if the user forgot his password. So we cannot ask him to login with it. 
This is why the route is ` NO_PERMISSION_REQUIRED`. No logged in requirement is needed in this case.
Note that this is the "completing the reset" part of the password-reset procedure. 
To trigger the reset itself, either an admin does it or the user during a "forgot my password" kind of login. The user would need to provide the valid username & email in that case, and the generated random token URL received at that email counter-validated with the email account.

magpie/alembic/versions/2021-01-04_954a9d7fe740_temporary_url_table.py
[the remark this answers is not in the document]
dbyrns: I agree and indeed if only resource_id has to be added in future that's not a huge deal given that we can also add foreign key on it. 
I was thinking to an operation requiring some kind of payload, like click here to apply this random permissions schema. And the schema must be store somehow so that the temporary url could apply it. For that kind of things, I was wondering if we could future proof the db by allowing a blob field where an operation could store whatever it want for its usage?

magpie/api/management/register/register_views.py
dbyrns: You're right! All of this make sense.

magpie/alembic/versions/2021-01-04_954a9d7fe740_temporary_url_table.py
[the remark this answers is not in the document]
fmigneault: I think it's better to wait for that specific use-case to happen and adjust the db with the most appropriate schema.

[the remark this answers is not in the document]
dbyrns: Correct!

dbyrns (approved):
Everything is good now and after discussion there is no need for a `create_temporary_url `function as everything in common is already  available in the `TemporaryToken `class.

fmigneault:
Relates to #161 and #378

PR #381, opened 2021-01-26
pin alembic to <1.5 to avoid failing parameters errors

fmigneault:


ChaamC:
There are still errors in the tests from github-actions. Should we try to fix them in this PR?

PR #383, opened 2021-01-27
Finish the missing tmp_url part for the webhooks

ChaamC:
I completed the missing part for the tmp_url with Francis' new code, and completed the FailingWebhook test case.

Note that I couldn't call the tmp_url directly from the webhook app, or at least, it would have been complicated, so instead, I just save the tmp_url value in the webhook_app, and retrieve it from the test case where I have easy access to the local Magpie app to call the tmp_url. I think this is fine as it still tests the tmp_url feature and tests the validity of the TemporaryToken id.

Let me know what you guys think!

### DIFF
--- a/magpie/api/management/user/user_utils.py
+++ b/magpie/api/management/user/user_utils.py
@@ -1,3 +1,4 @@
+import uuid
 from typing import TYPE_CHECKING
 
 import six
@@ -19,6 +20,7 @@ from ziggurat_foundations.models.services.user_resource_permission import UserRe
 from magpie import models
 from magpie.api import exception as ax
 from magpie.api import schemas as s
+from magpie.api.management.register.register_utils import TokenOperation
 from magpie.api.management.resource import resource_utils as ru
 from magpie.api.management.service.service_formats import format_service
 from magpie.api.management.user import user_formats as uf
@@ -119,11 +121,21 @@ def create_user(user_name, password, email, group_name, db_session):
 
     user_content = uf.format_user(new_user, new_user_groups)
 
+    # Create a token for the tmp_url, in case an error happens in the webhook services
+    token_id = uuid.uuid4()
+    webhook_token = models.TemporaryToken(
+        token=token_id,
+        operation=TokenOperation.WEBHOOK_ERROR.value,
+        user_id=new_user.id,
+        group_id=_get_group(group_name).id)
+    ax.evaluate_call(lambda: db_session.add(webhook_token), fallback=lambda: db_session.rollback(),
+                     http_error=HTTPForbidden, msg_on_fail=s.TemporaryToken_POST_ForbiddenResponseSchema.description)
+    tmp_url = webhook_token.url()
+
     # Force commit before sending the webhook requests, so that the user's status is editable if a webhook error occurs
     transaction.commit()
 
-    # Process any webhook requests
-    process_webhook_requests(WEBHOOK_CREATE_USER_ACTION, {"user_name": user_name, "tmp_url": "tmp_url:80/todo"}, True)
+    process_webhook_requests(WEBHOOK_CREATE_USER_ACTION, {"user_name": user_name, "tmp_url": tmp_url}, True)
 
     return ax.valid_http(http_success=HTTPCreated, detail=s.Users_POST_CreatedResponseSchema.description,
                          content={"user": user_content})
--- a/magpie/api/management/user/user_utils.py
+++ b/magpie/api/management/user/user_utils.py
@@ -135,7 +135,7 @@ def create_user(user_name, password, email, group_name, db_session):
     # Force commit before sending the webhook requests, so that the user's status is editable if a webhook error occurs
     transaction.commit()
 
-    process_webhook_requests(WebhookAction.WEBHOOK_CREATE_USER_ACTION,
+    process_webhook_requests(WebhookAction.CREATE_USER,
                              {"user_name": user_name, "tmp_url": tmp_url}, True)
 
     return ax.valid_http(http_success=HTTPCreated, detail=s.Users_POST_CreatedResponseSchema.description,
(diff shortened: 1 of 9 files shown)


magpie/api/management/user/user_utils.py
```diff
@@ -119,11 +121,21 @@ def _add_to_group(usr, grp):
 
     user_content = uf.format_user(new_user, new_user_groups)
 
+    # Create a token for the tmp_url, in case an error happens in the webhook services
+    token_id = uuid.uuid4()
+    webhook_token = models.TemporaryToken(
+        token=token_id,
```
fmigneault, line 127: I think the id can be omitted. The model should automatically handle its generation.

magpie/api/management/user/user_utils.py
```diff
@@ -119,11 +121,21 @@ def _add_to_group(usr, grp):
...
+    ax.evaluate_call(lambda: db_session.add(webhook_token), fallback=lambda: db_session.rollback(),
+                     http_error=HTTPForbidden, msg_on_fail=s.TemporaryToken_POST_ForbiddenResponseSchema.description)
+    tmp_url = webhook_token.url()
+
     # Force commit before sending the webhook requests, so that the user's status is editable if a webhook error occurs
     transaction.commit()
 
-    # Process any webhook requests
-    process_webhook_requests(WEBHOOK_CREATE_USER_ACTION, {"user_name": user_name, "tmp_url": "tmp_url:80/todo"}, True)
+    process_webhook_requests(WEBHOOK_CREATE_USER_ACTION, {"user_name": user_name, "tmp_url": tmp_url}, True)
```
fmigneault, line 138: Can you add the comment typings in this function?
I think the function should receive a `WebhookAction` enum, and do `.value` inside it.

tests/test_webhooks.py
```diff
@@ -217,6 +217,14 @@ def test_Webhook_CreateUser_FailingWebhook(self):
             users = utils.TestSetup.get_RegisteredUsersList(self)
             utils.check_val_is_in(self.test_user_name, users, msg="Test user should exist.")
 
+            # Check if the user's status is still set to 1, since the tmp_url has not been called yet
+            self.checkTestUserStatus(UserOKStatus)
+
+            # Retrieve the tmp_url and send the request to the magpie app
+            resp = requests.get(BASE_WEBHOOK_URL + "/get_tmp_url")
+            assert resp.text
```
fmigneault, line 225: use `utils.check_equal` with small message to say expected tmp url retrieval

tests/test_webhooks.py
```diff
@@ -10,6 +10,8 @@
 import tempfile
 import unittest
 from time import sleep
+from urllib.parse import urlparse
```
fmigneault, line 13: use `six.moves.urllib.parse`

magpie/api/management/user/user_utils.py
```diff
@@ -119,11 +121,21 @@ def _add_to_group(usr, grp):
 
     user_content = uf.format_user(new_user, new_user_groups)
 
+    # Create a token for the tmp_url, in case an error happens in the webhook services
+    token_id = uuid.uuid4()
+    webhook_token = models.TemporaryToken(
+        token=token_id,
+        operation=TokenOperation.WEBHOOK_ERROR.value,
```
fmigneault, line 128: I would rename `WEBHOOK_CREATE_USER` to make it distinct from other planned webhooks (on delete, on update)

magpie/models.py
```diff
@@ -443,7 +443,7 @@ def url(self, settings=None):
 
     def expired(self):
         expire = get_constant("MAGPIE_TOKEN_EXPIRE", raise_missing=False, raise_not_set=False, default_value=86400)
-        return (datetime.datetime.utcnow() - self.created) <= datetime.timedelta(seconds=expire)
+        return (datetime.datetime.utcnow() - self.created) > datetime.timedelta(seconds=expire)
```
fmigneault, line 446: Now that webhook test chain is complete, can we have a quick test for this?
Just use a small delta to ensure it triggers and validate the app returns the appropriate HTTP response.

fmigneault (changes requested):

magpie/api/management/user/user_utils.py
ChaamC: I did try this at first, but ran into problems. 
If I omit the `token=token_id,` when I create the TemporaryToken, I then try to retrieve the url with `tmp_url = webhook_token.url()` but I get `/tmp/None` as the token has not been generated yet.

I saw just above this code, for the user creation, I think the id also gets generated automatically and after creating a user and submitting it to the database, you retrieve the user again in order to have a 'refreshed" user with its automatically generated id.

Problem here I wasn't able to do this either. I can't retrieve the token since the other fields of the token are not unique (operation, user_id, group_id). I could possibly have multiple tokens that have the same values for those fields, so I would get multiple ids. 

There is also the by_token function which takes the token in parameter, but I can't use that either since it requires the automatically generated token id.
I didn't find another way to retrieve the 'refreshed' token with its automatically generated values.

So I will leave this as it is for now, but I am open to suggestions!

magpie/api/management/user/user_utils.py
ChaamC: I will change the WebhookAction list to use the ExtendedEnum (like the TokenOperation enum.), and will move the .value inside the function.

magpie/models.py
ChaamC: I am not sure to understand here what would be required.
Is it to test the webhook chain that uses a temporary token? 
There is the test `test_Webhook_CreateUser_FailingWebhook` which tests the token, by simulating an error in the webhook and calling the token's url back. It also uses the expired function in the background.

Or is it a new test I should add that tests the expired() function explicitly?

fmigneault: To speed up this PR, we will add it as another PR since not explicitly needed for the webhook feature.

magpie/api/webhooks.py
```diff
@@ -35,6 +27,14 @@
 LOGGER = get_logger(__name__)
 
 
+class WebhookAction(ExtendedEnum):
+    """
+    Actions supported by webhooks.
+    """
+    WEBHOOK_CREATE_USER_ACTION = "create_user"
+    WEBHOOK_DELETE_USER_ACTION = "delete_user"
```
fmigneault, line 35: Members should be only `CREATE_USER` and `DELETE_USER` since already within `WebhookAction` definition.

fmigneault (changes requested):
Only small rename, then good to go.

PR #399, opened 2021-03-24
Login improvements

### LINKED ISSUE #330: Help user find login route when proxy returns unauthorized
`WWW-Authenticate` header support needed in `MagpieAdapter` when cookies/auth header are missing.

relates to #255 basic auth support
relates to  bird-house/twitcher#96

### LINKED ISSUE #344: [Feature] Document authentication methods and related authenticated-user details
Although Magpie offers multiple external provider login methods, these are poorly documented, making it hard for devs and users to even know about them (relates to #173).

Also, even for "internal provider", it is not clearly documented how to authenticate one-self. Some details can be inferred from OpenAPI definitions, but one has to dig quite a lot to find such details. These details should be provided more explicitly in docs, such as : 
- how cookies are handled?
- what is the `auth_tkt` (or alternative value via config)?
- are Authentication Bearer token handled? (relates to #255)
- etc.

Given that authentication is such a main component of interactions with Magpie, appropriate details should be provided, in a dedicated section in the docs.

These details should also consider changes in #340 which provide new functionalities based on non-admin level users.
Doc should be provided to better explain contextual request behavior (ie: when `Logged User` is applicable to views).

fmigneault:
- Document Auth methods (fix #344)
- Provide WWW-Authenticate header (fix #330) 
- Forbid anonymous login (fix [DAC-388](https://www.crim.ca/jira/browse/DAC-388))
- Security improvements to login method


docs/authentication.rst
```diff
@@ -0,0 +1,280 @@
...
+    Terms :term:`Authentication` :term:`Providers`, :term:`External Providers` and :term:`External Providers` in this
+    chapter must not be confused with ``providers`` employed in :ref:`config_providers`. In this chapter, providers
+    refer to user-identity resolvers, in contrast to :term:`Service` definitions from the configuration files.
+
+.. _authn_requests:
+
+Authentication Requests
+---------------------------
+
+The most convenient way sign-in with `Magpie` is to employ the user interface provided on path
```
ChaamC, line 25: The most convenient way *to sign-in (missing "to" word)

magpie/utils.py
```diff
@@ -279,6 +281,53 @@ def convert_response(response):
...
+    if get_header("X-Requested-With", request.headers) == "XMLHttpRequest":
+        return None
+
+    # select error type: https://tools.ietf.org/html/rfc6750#section-3.1
+    if error_type not in ["invalid_token", "invalid_token", "insufficient_scope"]:
+        error_type = "invalid_token"
+    cookie_name = get_constant("MAGPIE_COOKIE_NAME", request)
+    magpie_url = get_magpie_url(request)
+    signin_url = "{}{}".format(magpie_url, SigninAPI.path)
+    login_url = "{}/ui/login".format(magpie_url)
```
ChaamC, line 314: Do we have a LoginAPI.path variable (like the SigninAPI.path found in the preceding line)?

tests/interfaces.py
```diff
@@ -408,6 +415,82 @@ def test_GetSession_InvalidTokens(self):
...
+        utils.check_or_try_logout_user(self)
+
+        path = s.SigninAPI.path
+        data_ok = {"user_name": self.test_user_name, "password": self.test_user_name}
+        data_bad = {"user_name": self.test_user_name, "password": "random-dont-care"}
+        for ctype in [CONTENT_TYPE_JSON, CONTENT_TYPE_HTML, CONTENT_TYPE_PLAIN]:
+            headers = {"Accept": ctype}
+            resp = utils.test_request(self, "GET", path, params=data_ok, headers=headers)
+            utils.check_response_basic_info(resp, 200, expected_method="GET", expected_type=ctype)
+
```
ChaamC, line 465: Why is it not required to add a logout before testing another sign in here?

ChaamC (commented):
LGTM in general. Great work here with multiple improvements/fixes to Magpie.

Just a few minor details or questions to check :)

magpie/utils.py
fmigneault: Sadly no. All UI paths don't have a service as they are not part of the OpenAPI.

tests/interfaces.py
fmigneault: Since login creates a cookie that overrides the previous one (same name/host), login again replaces it directly. 
I will add a logout for clarity of the intent, but it is not required.

fmigneault:
@ChaamC thanks again for your review, always spotting those typos :)

ChaamC (dismissed):

dbyrns (approved):
LGTM

PR #411, opened 2021-03-31
UI user status

fmigneault:
## changes

- Modify user status code to use an enum
- Adds an API param `status` to query users filtered by given status code
- Adds an UI icons that displays the user status
- Adds tests and utilities to validate the new features

*NOTE:*
it was easier to use #399 as base to avoid big diffs, consider review using to view only this PR changes : 
https://github.com/Ouranosinc/Magpie/compare/auth-doc-header...ui-user-status

## results

result in user list view 
image

result in user admin view / logged user own account view
image


dbyrns:
My first comment by looking at the screenshots is that it looks nice. My only concern is that I don't know what a green circle and a yellow triangle status means. I see two easy solutions (one of which may already be there) : Add a tooltip or in the details pane add a small text along the icon telling that "account is ready accross platform/some services failed to acknoledge this user". or something similar.

fmigneault:
@dbyrns 
When hovering the icons, it respectively shows `User ascount status is valid` or `User account status has an issue` (string in `title` of the HTML).

dbyrns (approved):
I was not sure that the "alt" text shows as a tooltip, but in this case LGTM!

fmigneault:
> I was not sure that the "alt" text shows as a tooltip

`alt` is to display something in case the image couldn't load or was not accessible for some reason

dbyrns:
> > I was not sure that the "alt" text shows as a tooltip
> 
> `alt` is to display something in case the image couldn't load or was not accessible for some reason

That was my first impression, but this is where the text came from when hovering over the icon, no? If not where does it come from?

fmigneault:
Field `title` here: 
https://github.com/Ouranosinc/Magpie/pull/411/files#diff-4ac24e7b273c7aee89de69aaef1c1d814031d4c1bb6d124ba321c162700bdf3eR182-R186

PR #412, opened 2021-04-02
Webhook and doc updates

fmigneault:
- prepare the field for other upcoming webhook implementations
- add more typing definitions for webhook items to detect invalid data containers
- provide more parameters to existing user create/delete webhook 
- provide multiple additional documentation details about existing webhooks, and generalize it for future ones incoming
- improve doc to use variable links 
    instead of plain ``<VAR>`` formatted text, an actual link to the definition is created to rapidly find it from anywhere in the doc

in preparation of [DAC-406](https://www.crim.ca/jira/browse/DAC-406) and [DAC-407](https://www.crim.ca/jira/browse/DAC-407)


magpie/typedefs.py
```diff
@@ -86,3 +87,12 @@
...
     ConfigList = List[ConfigItem]
     ConfigDict = Dict[Str, Union[Str, ConfigItem, ConfigList, JSON]]
+
+    WebhookTemplateParameters = Dict[Str, AnyValue]
+    WebhookConfigSettings = TypedDict("WebhookConfigSettings", {
+        "name": Str, "action": Str, "method": Str, "url": Str, "format": Str, "payload": JSON
+    })
+    WebhookLoadedSettings = TypedDict("WebhookLoadedSettings", {
+        "name": Str, "method": Str, "url": Str, "format": Str, "payload": JSON
+    })
```
ChaamC, line 97: Unsure here about the difference between WebhookConfigSettings and WebhookLoadedSettings, and the purpose of having 2 definition of settings?

ChaamC, line 97: Unsure here about the difference between WebhookConfigSettings and WebhookLoadedSettings, and the purpose of having 2 definition of settings?

fmigneault: The loaded settings don't have `action` compared to config definitions because of how the field was explicitly omitted by `setup_webhooks`, when they are regrouped by corresponding `action` under a list. These definitions detect that difference.

Was there any reason to omit the `action` in the first place other than because of redundant information?
If this is the case, we could leave it and simplify these types.

fmigneault: see other reply

ChaamC: I see why it was done in 2 definitions now. Yes, `action` was removed because of redundant information, since webhooks are reordered by action. 
I guess we have the choice between using these 2 type definitions, or letting the 'action' value in the webhook config setting, and just not using it in the processing later.

I wonder if using those 2 definitions could become confusing, although both solutions work.
What do you prefer?

magpie/api/webhooks.py
```diff
@@ -71,20 +92,20 @@ def replace_template(params, payload):
...
+        for template_param in params:
+            if template_param in WEBHOOK_TEMPLATE_PARAMS:
                 payload = payload.replace("{" + template_param + "}", params[template_param])
         return payload
     # For any other type, no replacing to do
     return payload
 
 
 def send_webhook_request(webhook_config, params, update_user_status_on_error=False):
+    # type: (WebhookConfigSettings, WebhookTemplateParameters, bool) -> None
```
ChaamC, line 108: If we keep both type definitions, I think it should be WebhookLoadedSettings here.

magpie/typedefs.py
fmigneault: I think I will leave the action in the dict even if redundant just to simplify everything and to align loaded config with whatever was in the original config. 
I was confused about the missing `action` in the loaded settings already until I realized this processing was accomplished.

ChaamC: Alright, that's good, I think it will reduce possible confusion here!

magpie/api/webhooks.py
```diff
@@ -4,65 +4,87 @@
...
+WEBHOOK_KEYS_OPTIONAL = {
+    "format"
+}
+WEBHOOK_KEYS = WEBHOOK_KEYS_REQUIRED | WEBHOOK_KEYS_OPTIONAL
 
-# These are the parameters permitted to use the template form in the webhook payload.
-WEBHOOK_TEMPLATE_PARAMS = ["user_name", "tmp_url"]
+# These are *potential* parameters permitted to use the template form in the webhook payload.
+# Each parameter transferred to any given webhook are provided distinctively for each case.
+WEBHOOK_TEMPLATE_PARAMS = ["user_name", "user_id", "user_email", "callback_url"]
```
dbyrns, line 44: Does user_id has a value for external use? Can it be used in the api? Since the primary ussage is in the database, if it is not exposed/expected in the API I would not provide it here neither.

dbyrns (approved):
I only have one comment, but overall it LGTM.

ChaamC (dismissed):

magpie/api/webhooks.py
fmigneault: I was thinking maybe it could (should?) be used by cowbird to generate for example the private working directories. 
To do things properly, definitions of user-specific items external to Magpie should be anonymized.

PR #414, opened 2021-04-07
Gunicorn update

### LINKED ISSUE #410: [Feature] Update gunicorn version
As flagged by pyup, gunicorn 20 addresses some (minor but potential) security issues. 
Must update to most recent version to get improvements. 
Requires changing command line calls (dockerfile and makefile) from `gunicorn` to `pserve` since INI port is not parsed anymore by `gunicorn` directly.

fmigneault:
- Update to most recent gunicorn version such that security fixes and performance improvements are integrated (resolve #410)
- General docs updates.

**important** 
needs to be properly updated in daccs/pavics, as the necessary binding changes using `pserve`
config INI with `host = localhost` does not work 
relates to https://github.com/bird-house/birdhouse-deploy/pull/143

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -43,4 +43,5 @@ COPY ./magpie $MAGPIE_DIR/magpie/
 RUN pip install --no-dependencies -e $MAGPIE_DIR
 
 # equivalent of `make cron start` without conda env
-CMD crond -c $CRON_DIR && gunicorn -b 0.0.0.0:2001 --paste $MAGPIE_CONFIG_DIR/magpie.ini --workers 10 --preload
+# bind to '0.0.0.0' such that any IP employed by the server will retrieve the application endpoint by default
+CMD crond -c $CRON_DIR && pserve $MAGPIE_CONFIG_DIR/magpie.ini bind=0.0.0.0:2001
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,7 +35,6 @@ RUN apk update \
     && apk --purge del .build-deps
 
 # install app package source, avoid copying the rest
-COPY ./bin $MAGPIE_DIR/bin/
 COPY ./config/magpie.ini $MAGPIE_CONFIG_DIR/magpie.ini
 COPY ./env/*.env.example $MAGPIE_ENV_DIR/
 COPY ./magpie $MAGPIE_DIR/magpie/
--- a/Dockerfile
+++ b/Dockerfile
@@ -43,4 +43,4 @@ RUN pip install --no-dependencies -e $MAGPIE_DIR
 
 # equivalent of `make cron start` without conda env
 # bind to '0.0.0.0' such that any IP employed by the server will retrieve the application endpoint by default
-CMD crond -c $CRON_DIR && pserve $MAGPIE_CONFIG_DIR/magpie.ini bind=0.0.0.0:2001
+CMD crond -c $CRON_DIR && pserve "$MAGPIE_CONFIG_DIR/magpie.ini"
--- a/config/magpie.ini
+++ b/config/magpie.ini
@@ -84,7 +84,7 @@ prefix = /magpie
 
 [server:main]
 use = egg:gunicorn#main
-host = localhost
+host = 0.0.0.0
 port = 2001
 timeout = 10
 workers = 3
--- a/config/magpie.ini
+++ b/config/magpie.ini
@@ -26,6 +26,9 @@ pyramid.includes =
 
 ## magpie
 # see all options in 'https://pavics-magpie.readthedocs.io/en/latest/configuration.html'
+# NOTE:
+#   below port definitions must be aligned with the value employed in [server:main] section if running locally
+#   when running within a server architecture, they should be updated accordingly to the desired server endpoint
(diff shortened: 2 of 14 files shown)


Dockerfile
```diff
@@ -35,12 +35,12 @@ RUN apk update \
...
 COPY ./config/magpie.ini $MAGPIE_CONFIG_DIR/magpie.ini
 COPY ./env/*.env.example $MAGPIE_ENV_DIR/
 COPY ./magpie $MAGPIE_DIR/magpie/
 # equivalent of `make install` without conda env and pre-installed packages
 RUN pip install --no-dependencies -e $MAGPIE_DIR
 
 # equivalent of `make cron start` without conda env
-CMD crond -c $CRON_DIR && gunicorn -b 0.0.0.0:2001 --paste $MAGPIE_CONFIG_DIR/magpie.ini --workers 10 --preload
+# bind to '0.0.0.0' such that any IP employed by the server will retrieve the application endpoint by default
+CMD crond -c $CRON_DIR && pserve "$MAGPIE_CONFIG_DIR/magpie.ini"
```
dbyrns, line 46: Number of workers was override to 10 here while set to 3 in the ini file.

config/magpie.ini
```diff
@@ -84,7 +84,7 @@ prefix = /magpie
 
 [server:main]
 use = egg:gunicorn#main
-host = localhost
+host = 0.0.0.0
 port = 2001
```
dbyrns, line 88: We have the now required environment variable MAGPIE_PORT and it is not used.

docs/usage.rst
```diff
@@ -123,12 +177,16 @@ are intended to validate older or pre-deployed servers, although with slightly m
 A basic `Magpie` instance can also be initialized by using the `docker-compose.yml.example`_ file.
 Note that the environment variables :envvar:`MAGPIE_TEST_REMOTE_SERVER_URL` and :envvar:`HOST_FQDN`
 must first be defined for the `remote` tests.
-For example, these default values should work with ``docker-compose``::
+For example, these default values should work with ``docker-compose``:
+
+.. code-block:: console
 
     export MAGPIE_TEST_REMOTE_SERVER_URL=http://localhost:2001
```
dbyrns, line 184: Is it still valid since we also need the MAGPIE_PORT variable?

dbyrns (dismissed):
Some minor quirks to be answered, but good overall.

Dockerfile
fmigneault: It should now be done in the INI. It didn't make much sense to be in the command in the first place. 
It is very hard to track that this value was overridden here, and the first impression is usally that whatever was in the INI is the 'truth'.

config/magpie.ini
fmigneault: Yes. I have yet to find a way to simplify this. I have tried defining a "shared" section where `magpie.port`, and `[server:main]`'s `port` would use the same reference, but it was not working.

docs/usage.rst
fmigneault: This one is only for tests. 
`MAGPIE_PORT` is needed only if `MAGPIE_URL` or `magpie.url` in the INI was not defined.

Dockerfile
dbyrns: Yeah, I'm fine with the fact that the INI is the 'truth'. I'm more concerned about the fact that we go from 10 to 3 workers. I would not have said a word if the INI set workers to 10 too.

config/magpie.ini
dbyrns: Ok, then I suggest that you add a small comment telling users that port must be defined at both place.

Dockerfile
fmigneault: I can set it by default to 10, so that it will be transparent for anyone that used the image without override of the command.

config/magpie.ini
fmigneault: I still wonder though why `MAGPIE_PORT` even needs to be defined in your case ?
It is important to remember that ``MAGPIE_PORT``, ``MAGPIE_URL``, etc.  are only variable to set the "exposed" URL. 
So an application running on `http://my-pavics-server.ca` should define exactly that value in `MAGPIE_URL` only. 
The ``port`` of ``[server:main]`` still remains `2001` because it is where the local app is accessible from the docker. 

The ``MAGPIE_PORT`` should only be `2001` when running locally.

Dockerfile
dbyrns: Yes that's the kind of backward compatibility that we must keep in mind and avoid unnecessary changes when possible.

config/magpie.ini
dbyrns: From my perspective I wonder why you did not saw this error on your side 😜, I guess that MAGPIE_URL is not set properly somewhere. But with https://github.com/Ouranosinc/Magpie/pull/418 I'm not worried anymore. 
I also suggest that you remove MAGPIE_PORT and fix MAGPIE_URL if required in https://github.com/bird-house/birdhouse-deploy/pull/143 before running tests.

dbyrns (approved):
Everything is set. Thank you.

PR #419, opened 2021-04-09
Webhook permissions

fmigneault:
fix [DAC-407](https://www.crim.ca/jira/browse/DAC-407)

- add webhooks that can be triggered when any combination of (user|group, service|resource, permission, created|deleted) occurs. 
- adjust the format of webhook template parameters to avoid silently failing errors due to YAML parsing method


magpie/api/webhooks.py
```diff
@@ -60,27 +73,81 @@ class WebhookAction(ExtendedEnum):
...
     UPDATE_USER_STATUS = "update_user_status"
     """
-    Triggered when an existing user status gets successfully updated.
+    Triggered when an existing :term:`User` status gets successfully updated.
 
     .. seealso::
         :ref:`webhook_user_update_status`
     """
 
+    CREATE_USER_PERMISSION = "create_user_permission"
```
dbyrns, line 97: Should we have an UPDATE_(USER|GROUP)_PERMISSION, or it's resolve as a delete then create operations?

magpie/api/webhooks.py
```diff
@@ -188,8 +257,8 @@ def send_webhook_request(webhook_config, params, update_user_status_on_error=Fal
     except Exception as exception:
         LOGGER.error("An exception has occurred with the webhook request : %s", webhook_config["name"])
         LOGGER.error(str(exception))
-        if "user_name" in params and update_user_status_on_error:
-            webhook_update_error_status(params["user_name"])
+        if "user.name" in params and update_user_status_on_error:
```
dbyrns, line 260: Hum, since we have group permission hook now, does this mean that we should have error status for groups too?
On the other side, having a permission webhook error does not mean that the user is "on error". The status was really intended for the creation webhook.
But then how to signal that permission webhooks are malfunctionning?

dbyrns (commented):
2 Questions. Hoping it's not too much trouble.

magpie/api/webhooks.py
fmigneault: Yes its handled as 2 separate operations, so you can subscribe to only the need one(s).
I feel if we add an update one, we would have weird situations where you could have simultaneous create+update+delete configs, which would result in 4 webhook triggers.

dbyrns: It's fine, I was only wondering if via an hypothetic update API call, we could mess a synched service.

magpie/api/webhooks.py
dbyrns: > We could have that for group also, but for the moment, even the user permission change does not set a user-status error. Only the user create/update webhook change the status to error.
I think we would need a status on specific permissions, because as you said, setting an error status for the user or the whole group feels likes the "wrong target".

Yep I agree with the current state of things. 

> Another issue is how to report all those individual permissions status error in a useful way? Aggregate too much, and we don't know what is erroneous. Too granular, and we can never go through everything.

If I check the Magpie logs, will I get the required granularity to pinpoint the issue? If no, maybe you could log a little bit more on those issues. 
Then for the big picture, could we aggregate issues by webhook config? So that at least an admin, when logged, could see that some external service is misbehaving and be hinted that he should check the logs? (Not for this PR, but you could open a nice to have issue for Magpie)

magpie/api/webhooks.py
fmigneault: I think you are right, we can do a PUT request to replace any `access` or `scope` value for a target permission.
We can swap `allow` to `deny` or grant/remove lower level resources some permissions. 
So indeed, a third "on update" hook would be needed here.

magpie/api/webhooks.py
fmigneault: I think most of the errors would be hard to retrieve from logs, since part of the operation happens outside of magpie (the middleware subscribed in the webhook). For now, there is probably not even enough log details, but if it did, it could probably only say that the webhook replied with an error. 

What we could do is register the webhook failures in the db with relevant actions and configs. 
Those could be returned in a list, maybe filtered "by issue" similarly to how sentry does. I'm still not sure how granular per error cause would produce the best result.

dbyrns: > only say that the webhook replied with an error

For me it's enough on the Magpie side. If Magpie can tell us that "service X" caused an error while creating a user, then we will know where to search.

I don't think we need a database for this right now. We don't need another sentry. Rather than going that route I would prefer if we could silently raise an error to sentry. But again this is out of scope for this PR.

fmigneault: In that case, it is logged. It provides the `name` of the failing webhook and the "reason" of the http-error.

PR #430, opened 2021-05-06
Fix config loading

### LINKED ISSUE #429: [BUG] Error parsing extra details in some startup config files
**Describe the bug**

Config format is incorrectly parsed (list of dict instead of directly the dict) when additional sections are provided specifically in `permissions.cfg` or `providers.cfg` files. This doesn't seem to be the case in `config.yml` files, but should be validated with corresponding tests.

**To Reproduce**
1. Load `permissions.cfg` file that has both `permissions` and another section (eg: `users`)
2. Error logged

**Expected behavior**

**Desktop (please complete the following information):**
 - Version: seen in 3.8, but could be anything before.

**Additional Details**

``` log
Traceback (most recent call last):
  File "/usr/local/bin/gunicorn", line 8, in <module>
    sys.exit(run())
  File "/usr/local/lib/python3.7/site-packages/gunicorn/app/wsgiapp.py", line 61, in run
    WSGIApplication("%(prog)s [OPTIONS] [APP_MODULE]").run()
  File "/usr/local/lib/python3.7/site-packages/gunicorn/app/base.py", line 223, in run
    super(Application, self).run()
  File "/usr/local/lib/python3.7/site-packages/gunicorn/app/base.py", line 72, in run
    Arbiter(self).run()
  File "/usr/local/lib/python3.7/site-packages/gunicorn/arbiter.py", line 60, in __init__
    self.setup(app)
  File "/usr/local/lib/python3.7/site-packages/gunicorn/arbiter.py", line 120, in setup
    self.app.wsgi()
  File "/usr/local/lib/python3.7/site-packages/gunicorn/app/base.py", line 67, in wsgi
    self.callable = self.load()
  File "/usr/local/lib/python3.7/site-packages/gunicorn/app/wsgiapp.py", line 50, in load
    return self.load_pasteapp()
  File "/usr/local/lib/python3.7/site-packages/gunicorn/app/wsgiapp.py", line 46, in load_pasteapp
    return load_pasteapp(self.cfgurl, self.relpath, global_conf=self.cfg.paste_global_conf)
  File "/usr/local/lib/python3.7/site-packages/gunicorn/app/pasterapp.py", line 71, in load_pasteapp
    global_conf=global_conf)
  File "/usr/local/lib/python3.7/site-packages/paste/deploy/loadwsgi.py", line 253, in loadapp
    return loadobj(APP, uri, name=name, **kw)
  File "/usr/local/lib/python3.7/site-packages/paste/deploy/loadwsgi.py", line 278, in loadobj
    return context.create()
  File "/usr/local/lib/python3.7/site-packages/paste/deploy/loadwsgi.py", line 715, in create
    return self.object_type.invoke(self)
  File "/usr/local/lib/python3.7/site-packages/paste/deploy/loadwsgi.py", line 150, in invoke
    **context.local_conf)
  File "/usr/local/lib/python3.7/site-packages/paste/deploy/util.py", line 55, in fix_call
    val = callable(*args, **kw)
  File "/usr/local/lib/python3.7/site-packages/paste/urlmap.py", line 31, in urlmap_factory
    app = loader.get_app(app_name, global_conf=global_conf)
  File "/usr/local/lib/python3.7/site-packages/paste/deploy/loadwsgi.py", line 356, in get_app
    name=name, global_conf=global_conf).create()
  File "/usr/local/lib/python3.7/site-packages/paste/deploy/loadwsgi.py", line 715, in create
    return self.object_type.invoke(self)
  File "/usr/local/lib/python3.7/site-packages/paste/deploy/loadwsgi.py", line 235, in invoke
    filtered = context.next_context.create()
  File "/usr/local/lib/python3.7/site-packages/paste/deploy/loadwsgi.py", line 715, in create
    return self.object_type.invoke(self)
  File "/usr/local/lib/python3.7/site-packages/paste/deploy/loadwsgi.py", line 152, in invoke
    return fix_call(context.object, context.global_conf, **context.local_conf)
  File "/usr/local/lib/python3.7/site-packages/paste/deploy/util.py", line 55, in fix_call
    val = callable(*args, **kw)
  File "/opt/local/src/magpie/magpie/app.py", line 75, in main
    magpie_register_permissions_from_config(perm_cfg, db_session=db_session)
  File "/opt/local/src/magpie/magpie/register.py", line 944, in magpie_register_permissions_from_config
    _process_permissions(perms, magpie_url, cookies_or_session, users, groups)
  File "/opt/local/src/magpie/magpie/register.py", line 974, in _process_permissions
    users_conf = _make_config_registry(users, "username")
  File "/opt/local/src/magpie/magpie/register.py", line 959, in _make_config_registry
    cfg_key = cfg.get(key, None)
AttributeError: 'list' object has no attribute 'get'
```
using this config
https://github.com/bird-house/birdhouse-deploy/commit/6bd69b17f62b92e4e15e387456d95be8d3bca2c0

fmigneault:
# Features / Changes
* Add explicit typing definitions of configuration files and resolved settings to facilitate discovery of invalid
  handling of formats or parameters during parsing and startup registration.
* Apply many documentation updates in both configuration sections and the corresponding configuration example headers.

# Bug Fixes
* Fix ``users`` and ``groups`` registration configurations not respecting update method when conflicting
  definitions occur. They will respect alphabetical file name order and later ones remain.
* Fix ``users`` and ``groups`` registration configurations not correctly parsed when multiple files where employed (fixes #429).

### DIFF
--- a/config/config.yml
+++ b/config/config.yml
@@ -1,5 +1,8 @@
-## Combined configuration definition to setup objects at Magpie startup
-## See documentation 'Configuration'
+## Combined Configuration
+## ------------------------
+## Prepare or suppress known entity definitions at Magpie startup.
+## See documentation 'Configuration' for detail about loading method, order and other relevant information.
+## ------------------------
 
 # All following sections are optional.
 # All parameters must fulfill data validation rules as per corresponding API requests.
@@ -13,8 +16,8 @@ providers:
 permissions:
   # see full format details in 'permissions.cfg'
   # for each definition:
-  #   can refer to users and groups that will be created dynamically as needed with corresponding information
-  #   plain user or group name not matched against below entry will create it with defaults for other fields
+  #   can refer to below 'users' and 'groups' that will be created dynamically as needed with corresponding information
+  #   user or group name not matched against any extended definition will be created with defaults for other fields
 
 # extended definitions for users to create
 # default values will be used for missing optional fields
@@ -22,9 +25,9 @@ permissions:
 #   will default to only using the user name and defaults for every other field
 users:
   - username: username  # required if entry provided
-    password:           # optional
-    email:              # optional
-    group: <groupname>  # optional (either one of the groups, or plain string to generate it with defaults)
+    password:           # optional (default: random password of size MAGPIE_PASSWORD_MIN_LENGTH)
+    email:              # optional (default: <username>@mail.com)
+    group: <groupname>  # optional (one of 'groups' entry, or plain string to generate it with defaults, else anonymous)
 
 # extended definitions for groups to create
 # default values will be used for missing optional fields
@@ -32,8 +35,8 @@ users:
 #   will default to only using the group name and defaults for every other field
 groups:
   - name: <groupname>                 # required if entry provided
-    description: <some description>   # optional
-    discoverable: True                # optional
+    description: <some description>   # optional (default: empty)
+    discoverable: True                # optional (default: False)
 
 # Definitions of all the webhooks urls that will be called when creating or deleting a users.
 webhooks:
--- a/config/config.yml
+++ b/config/config.yml
@@ -27,7 +27,8 @@ users:
   - username: username  # required if entry provided
     password:           # optional (default: random password of size MAGPIE_PASSWORD_MIN_LENGTH)
     email:              # optional (default: <username>@mail.com)
-    group: <groupname>  # optional (one of 'groups' entry, or plain string to generate it with defaults, else anonymous)
+    group: <group.name> # optional (default: anonymous, otherwise one of 'groups' matched by 'name')
+                        #          (if group name doesn't exist in 'groups', other parameters use defaults)
 
 # extended definitions for groups to create
 # default values will be used for missing optional fields
--- a/config/permissions.cfg
+++ b/config/permissions.cfg
@@ -25,12 +25,18 @@
 #   Both can also be specified simultaneously to create/remove the same permission for both cases.
 #
 #   Whenever a specified user/group doesn't exist in the database for permission creation, it gets created dynamically.
-#   When this dynamic creation must occur, sections 'users' and 'groups' are looked for possible extended definitions,
-#   as presented in 'config.yml'. Otherwise, the corresponding defaults are employed when no match was found.
+#   When this dynamic creation must occur, sections 'users' and 'groups' are looked for possible extended definitions to
+#   obtain the complete set of parameters applicable for each creation.
 #
-#   The 'users' and 'groups' sections must be at the top-level of the configuration (same as 'permissions').
-#   This is to allow reuse of the same named reference multiple times across many permissions operations without
-#   needing to repeat the whole user/group parameters each time.
+#   Lookup of 'users' and 'groups' is done transversally over all loaded configuration files within the configured
+#   directory. This means that 'users' could be defined in one file, and referenced in another. Note that conflicting
+#   items will override each other according to alphabetical load order of those files.
+#   When no match can be found in any if the respective sections, default parameters of those definitions are employed.
+#
+#   The 'users' and 'groups' sections must be at the top-level of configuration files (same level as 'permissions').
+#   This is to allow reuse of the same named reference multiple times across many 'permissions' items without
+#   needing to repeat the whole set of parameters for corresponding user/group each time.
+#   Examples of such sections are presented in 'config.yml'.
 #
 # Permission:
 # -----------
--- a/config/permissions.cfg
+++ b/config/permissions.cfg
@@ -1,6 +1,7 @@
 # Creates Magpie services/resource permissions with specified configuration values on application startup.
 #
 #   This configuration is parsed after 'providers.cfg' (if found) to allow referencing existing service definitions.
+#   See documentation 'Configuration' for more detail about loading method, order and other relevant information.
 #
 #   - Default location is 'magpie/config/permissions.cfg'.
 #   - Directory location can be overridden with 'MAGPIE_CONFIG_DIR' (using filename 'permissions.cfg')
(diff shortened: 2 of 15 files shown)


config/config.yml
```diff
@@ -13,27 +16,27 @@ providers:
...
 #   not providing an entry referenced somewhere else (e.g.: in permissions)
 #   will default to only using the user name and defaults for every other field
 users:
   - username: username  # required if entry provided
-    password:           # optional
-    email:              # optional
-    group: <groupname>  # optional (either one of the groups, or plain string to generate it with defaults)
+    password:           # optional (default: random password of size MAGPIE_PASSWORD_MIN_LENGTH)
+    email:              # optional (default: <username>@mail.com)
+    group: <groupname>  # optional (one of 'groups' entry, or plain string to generate it with defaults, else anonymous)
```
MatProv, line 30: Reverse order : 'optional (default: "anonymous", one of ..)'

Why is "<groupname>"'s there but no "<email>"? Either remove or add missing placeholders.

config/permissions.cfg
```diff
@@ -11,12 +12,26 @@
...
+#   At least one of [user, group] must be provided for every permission entry specified to create/remove.
+#   Both can also be specified simultaneously to create/remove the same permission for both cases.
+#
+#   Whenever a specified user/group doesn't exist in the database for permission creation, it gets created dynamically.
+#   When this dynamic creation must occur, sections 'users' and 'groups' are looked for possible extended definitions,
+#   as presented in 'config.yml'. Otherwise, the corresponding defaults are employed when no match was found.
+#
+#   The 'users' and 'groups' sections must be at the top-level of the configuration (same as 'permissions').
+#   This is to allow reuse of the same named reference multiple times across many permissions operations without
+#   needing to repeat the whole user/group parameters each time.
```
MatProv, line 33: And what happens if it's not the case? Simply repeat the user/group parameters each time, or might have undesired behaviour?

fmigneault: It merges all `users` and all `groups` in every matched config. 
Then, using those mappings, attempts to resolve corresponding `user` and group` references. 
If matched, it uses the full definition. Otherwise, the plain string for `name` + other fields defaults. 

If this definition clear enough or needs more details? I will add is as is if it helps understand.

config/config.yml
fmigneault: +1 for the reverse order

I don't understand the other comment about the `""`

docs/configuration.rst
```diff
@@ -229,17 +231,29 @@ These settings can be used to specify where to find other settings through custo
...
-        (depending on OS implementation). Duplicate entries could therefore be loaded in inconsistent order.
-        Please refer to `providers.cfg`_ for specific format details and loading methodology according to arguments.
+        If a directory path is specified, the order of loaded configuration files is alphabetical.
+        Matching :term:`Services` will be overridden by files loaded last.
+
+    .. versionchanged:: 1.7.4
+
+        Loading order of multiple files was **NOT** guaranteed prior to this version.
+
+        This could lead to some entries to be loaded in inconsistent order.
```
MatProv, line 256: Good to know! Thanks for specifying

config/config.yml
MatProv: See the edit in previous comment

config/permissions.cfg
MatProv: The explanation makes sense, but maybe just put the "otherwise" section after the phrase that states "must". 
Usually when reading something like "there must be", we seek to find a "and otherwise, this would happen".
If it's only to avoid repeating the user/group params each time, maybe just state that it would be the only consequence.

docs/configuration.rst
```diff
@@ -248,31 +262,45 @@ These settings can be used to specify where to find other settings through custo
...
+
+    .. versionchanged:: 3.6
+        The configuration can also contain a ``webhooks`` section, as described in :ref:`config_webhook` and
+        presented in the sample :ref:`config_file`.
+
+    .. warning::
+        When this setting is defined, all other combinations of :envvar:`MAGPIE_CONFIG_DIR`,
+        :envvar:`MAGPIE_PERMISSIONS_CONFIG_PATH` and :envvar:`MAGPIE_PROVIDERS_CONFIG_PATH` are effectively
+        ignored in favour of definitions in this file. It is not possible to employ the :ref:`config_file` at
+        the same time as multi-configuration loading from a directory.
```
MatProv, line 295: Haven't looked but I guess that there is a check to know if user tries to load `config_file` and multi-configuration at the same time?

config/permissions.cfg
fmigneault: I pushed another update, let me know if it is clearer.

magpie/api/webhooks.py
```diff
@@ -284,14 +284,14 @@ def setup_webhooks(config_path, settings):
     """
 
     settings["webhooks"] = defaultdict(lambda: [])
-    webhooks_conf = settings["webhooks"]  # type: WebhookConfig
+    webhooks_conf = settings["webhooks"]  # type: WebhookSettings
```
MatProv, line 287: Be careful with this naming. In `send_webhook_request()` `webhook_config` is of type `WebhookConfigItem`, while here `webhooks_conf` is of type `WebhookSettings`.

To avoid confusion, maybe use `webhooks_settings_config` here, instead.

docs/configuration.rst
fmigneault: As soon as you define `MAGPIE_CONFIG_PATH`, you enforce the single-file mode. 
Otherwise, it attempts to resolve according to `MAGPIE_CONFIG_DIR`, `MAGPIE_PERMISSIONS_CONFIG_PATH` and `MAGPIE_PROVIDERS_CONFIG_PATH`. 

It is more detailed in their specific `:envvar:` section because it is repeating itself a lot otherwise and harder to maintain the docs accurate everywhere. I don't re-read everything all times, so I try to reference existing explanations instead of repeating to avoid discrepancies.

magpie/api/webhooks.py
fmigneault: This is why I added the typings. 
Different devs using interchangeably the names for config/setttings. 

But I agree name can be more explicit also.

MatProv: Yes of course the choice of config or settings refers to the same concept. 
What's important is the consistence, with the variable naming/type, wether it's called a config or a setting.

config/config.yml
```diff
@@ -13,27 +16,28 @@ providers:
...
 # extended definitions for users to create
 # default values will be used for missing optional fields
 #   not providing an entry referenced somewhere else (e.g.: in permissions)
 #   will default to only using the user name and defaults for every other field
 users:
   - username: username  # required if entry provided
-    password:           # optional
-    email:              # optional
-    group: <groupname>  # optional (either one of the groups, or plain string to generate it with defaults)
+    password:           # optional (default: random password of size MAGPIE_PASSWORD_MIN_LENGTH)
```
dbyrns, line 28: Just wondering why this is optional as if a random password is used, nobody can access it and thus the whole account is useless.

dbyrns (approved):
It looks good overall, except for my only comment below, but since @MatProv is doing a thorough review I let him do the final approval.

config/permissions.cfg
MatProv: Yes, thanks for the update.

docs/configuration.rst
MatProv: Ok then if it's specified in the envvar part no need to repeat over here.
The goal was to avoid confusion in case someone tries both config setups and ends up with the `config_file` one.
But since it seems specified in envvar, that looks good to me.

config/config.yml
```diff
@@ -27,7 +27,8 @@ users:
   - username: username  # required if entry provided
     password:           # optional (default: random password of size MAGPIE_PASSWORD_MIN_LENGTH)
     email:              # optional (default: <username>@mail.com)
-    group: <groupname>  # optional (one of 'groups' entry, or plain string to generate it with defaults, else anonymous)
+    group: <group.name> # optional (default: anonymous, otherwise one of 'groups' matched by 'name')
```
MatProv, line 30: Well here my previous question wasn't related to "groupname" vs "group.name".
I was wondering about why this "groupname" is specified, but not, for example, "email" or "password" in the previous lines.

MatProv (approved):
Minor things to change, see comments, but overall looks good to me.

config/config.yml
fmigneault: Since the admin can override, it is not critical to be missing the information. 
Also, without this default, there is no way to simply have the username string everywhere else that `user` is specified. 

Having the default allows to quickly define a few users for testing purposes. 
Obviously, for users that would be used for actual operations, the password would have to be provided in the definition.

config/config.yml
fmigneault: Ah, I see. It is just to make it more obvious it if provided, it can be a reference value that is equal to a name from the `groups` section, such that more parameters can be found. So, to make illustrate it is more than just a plain string.

PR #431, opened 2021-05-06
Webhook config multi

fmigneault:
Identified while working on #430 

This will be of good interest when integrating with `cowbird`. 

## main change
* Add ``MAGPIE_WEBHOOKS_CONFIG_PATH`` configuration setting / environment variable that allows potentially using
  multiple configuration files for `Webhooks`. This parameter is notably important for developers that where using the
  ``MAGPIE_PROVIDERS_CONFIG_PATH`` or ``MAGPIE_PERMISSIONS_CONFIG_PATH`` settings to load multiple files, as they
  cannot be combined with single configuration provided by ``MAGPIE_CONFIG_PATH``, which was the only supported way to
  provide `Webhooks` definitions.

Other updates are related to above.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -188,6 +188,10 @@ field.
     The ``webhook`` section allows to define external connectors to which `Magpie` should send requests following
     certain events. These are described in further details in :ref:`config_webhook` section.
 
+.. versionadded:: 3.12
+    Variable :envvar:`MAGPIE_WEBHOOKS_CONFIG_PATH` was added and will act in a similar fashion as their providers and
+    permissions counterparts, to load definitions from multiple configuration files.
+
 
 .. _config_constants:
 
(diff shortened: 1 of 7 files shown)


docs/configuration.rst
```diff
@@ -290,9 +318,10 @@ These settings can be used to specify where to find other settings through custo
 
     .. warning::
         When this setting is defined, all other combinations of :envvar:`MAGPIE_CONFIG_DIR`,
-        :envvar:`MAGPIE_PERMISSIONS_CONFIG_PATH` and :envvar:`MAGPIE_PROVIDERS_CONFIG_PATH` are effectively
-        ignored in favour of definitions in this file. It is not possible to employ the :ref:`config_file` at
-        the same time as multi-configuration loading from a directory.
+        :envvar:`MAGPIE_PERMISSIONS_CONFIG_PATH`, :envvar:`MAGPIE_PROVIDERS_CONFIG_PATH` and
+        :envvar:`MAGPIE_WEBHOOKS_CONFIG_PATH` are effectively ignored in favour of definitions in this file.
```
MatProv, line 322: Nice to have this specified in the config. 
Do you think it would be helpful to also have a warning in case one of these envvars are non-null while having `MAGPIE_CONFIG_PATH` defined?

dbyrns (approved):

MatProv (approved):

docs/configuration.rst
fmigneault: Yeah, we can add a log entry that says var "x" not used because `MAGPIE_CONFIG_PATH` was detected.

PR #435, opened 2021-06-11
User self-registration

fmigneault:
# Features / Changes

* Changed ``UserStatuses.WebhookErrorStatus = 0`` to ``UserStatuses.WebhookError = 2`` to provide further
  functionalities. Migration script applies this change to existing ``User`` entries.
* Changed the returned ``status`` value by the API routes to use the string name representation instead of the integer.
* Changed ``status`` search query handling of ``GET /users`` path for improved search and filtering capabilities.
* Add new ``UserStatuses.Pending = 4`` value that can be queried by administrators.
* Add ``UserPending`` object with corresponding table for pending approval by an administrator for some new
  self-registered user. Migration script creates the table with expected fields.
* Add new requests under ``/register/users`` and ``/ui/register/users`` endpoints for user account self-registration.
* Add UI view to display pending user registration details.
* Add UI icon to indicate when a listed user is pending registration approval or email validation.
* Disable user email self-update (when not administrator) both on the API and UI side
  whenever ``MAGPIE_USER_REGISTRATION_ENABLED`` was activated to avoid losing the confirmation of the original email.
* Add configuration setting ``MAGPIE_USER_REGISTRATION_ENABLED`` to control whether user account self-registration
  feature should be employed.
  With it comes multiple other ``MAGPIE_USER_REGISTRATION_<...>`` settings to customize notification emails.
* Add multiple ``MAGPIE_SMTP_<...>`` configuration settings to control connections to notification email SMTP server.
* Add ``empty_missing`` flag to ``get_constant`` utility to allow validation against existing environment variables or
  settings that should be considered as invalid when resolved value is an empty string.
* Add missing ``format`` for applicable ``url`` and ``email`` elements in the OpenAPI specification.
* Add better logging options control in CLI operations.
* Add new CLI helper ``send_email`` to test various email template generation and SMTP configurations to send emails.
* Replace ``-d`` option of ``register_providers`` CLI operation (previously used to select database mode)
  by ``--db`` to avoid conflict with logging flags.
* Replace ``-d`` and ``-l`` options of ``batch_update_users`` CLI operation respectively by ``-D`` and ``-L``
  to avoid conflict with logging flags.

# Bug Fixes
* Explicitly disallow duplicate email entries, both with pre-validation and literal database values.
    Note that any duplicate email will be raised an migration script will fail. Manual cleanup of the undesired entry
    will be required, as `Magpie` will not be able to assume which one corresponds to the valid user to preserve.

* Add ``ziggurat_foundations`` extensions for Pyramid directly in the code during application setup such that an INI
  configuration file that omits them from ``pyramid.include`` won't cause `Magpie` to break.

# Other references
Implements feature [DAC-117](https://www.crim.ca/jira/browse/DAC-117)


fmigneault:
@ChaamC 
Tagging you as some of the files/configs needed for T&C feature are added here.

config/magpie.ini
```diff
@@ -64,15 +82,6 @@ cache.acl.enabled = false
 cache.service.enable = false
 # cache.service.expire = 10
 
-# ziggurat
```
dbyrns, line 67: It looks like this config is now loaded dynamically. Not so sure where this config should be put as they're is no config file containing them anymore. Maybe a little message here telling admins where they should configure ziggurat from now on.

docs/authentication.rst
```diff
@@ -280,3 +288,130 @@ to :term:`User` identification against local accounts in `Magpie` using :envvar:
...
+
+    2. Upon reception of a valid new registration, an email will be sent to the submitted email such that the
+       :term:`Pending User` can validate its address with the provided confirmation link. This step is
+       always required.
+
+    3. Once the validation is received, there are two possible outcomes according to configured value of
+       :envvar:`MAGPIE_USER_REGISTRATION_APPROVAL_ENABLED`.
+
+        A) When :envvar:`MAGPIE_USER_REGISTRATION_APPROVAL_ENABLED` is **NOT** activated, only email validation is
+           required from the confirmation email. Receiving the confirmation in that case immediately completes
```
dbyrns, line 345: The "Receiving" term is not clear here. User receiving the confirmation email is not enough, it must follow the confirmation link.

dbyrns (commented):
Review in progress 5/74

config/magpie.ini
fmigneault: This is now included directly in the code statically instead of dynamically by INI loaded when starting Magpie becausee a config that did not provide them simply failed to work as intended. It is something that cannot be configured. 
For someone not working actively with Magpie, it is very hard to figure out how that could be the cause of error when it was omitted by error.

docs/authentication.rst
fmigneault: see updated wording

magpie/alembic/versions/2021-04-16_dea413e13a8a_pending_users_and_statuses.py
```diff
@@ -0,0 +1,47 @@
...
+                              default=datetime.datetime.utcnow, server_default=sa.func.now())
+                    )
+
+    session = Session(bind=op.get_bind())
+    query = sa.update(User).where(User.c.status == 0).values(status=2)  # use literals to avoid enum value changes
+    session.execute(query)
+    session.commit()
+
+
+def downgrade():
```
dbyrns, line 46: Does status 2 should be bring back to 0?

magpie/api/management/register/register_utils.py
```diff
@@ -1,50 +1,110 @@
...
+    elif tmp_token.operation == TokenOperation.USER_REGISTRATION_CONFIRM_EMAIL:
+        LOGGER.debug("[User Registration - Step 3] received email validation for user: [%s]", tmp_token.user.user_name)
+        response = handle_user_registration_confirmation(tmp_token, request)
+
+    # User Registration Procedure - Step (4): reception of the administrator approve/decline URL
+    elif tmp_token.operation in [TokenOperation.USER_REGISTRATION_ADMIN_APPROVE,
+                                 TokenOperation.USER_REGISTRATION_ADMIN_DECLINE]:
+        ax.verify_param(ar.has_admin_access(request), is_true=True, with_param=False,
+                        http_error=HTTPForbidden, msg_on_fail=s.HTTPForbiddenResponseSchema.description)
+        LOGGER.debug("[User Registration - Step 4] admin reviewed pending user: [%s]", tmp_token.user.user_name)
```
dbyrns, line 91: Should we log the admin user approving or declining the user?

magpie/api/management/register/register_utils.py
```diff
@@ -79,3 +139,219 @@ def get_discoverable_group_by_name(group_name, db_session):
...
+    """
+    require_approve = asbool(get_constant("MAGPIE_USER_REGISTRATION_APPROVAL_ENABLED", request, default_value=False,
+                                          print_missing=True, raise_missing=False, raise_not_set=False))
+    if not require_approve:
+        LOGGER.debug("[User Registration - Step 3A] moving on to complete registration (no approval required)")
+        complete_user_registration(tmp_token, request)
+        data = {
+            # from the message UI page template, it is possible to see the 'Login' button in the Magpie header
+            # user should be able to find its way there to use its registered account
+            "message": cleandoc("""
```
dbyrns, line 230: Why the message is not part of the mako template?

dbyrns (commented):
Review in progress 19/74

magpie/models.py
```diff
@@ -132,14 +145,309 @@ def __str__(self):
...
+
+    # nothing to do, only improve typing to avoid complaints about expecting 'int'
+    def __xor__(self, other):  # type: (Union[UserStatuses, int]) -> UserStatuses
+        return super(UserStatuses, self).__xor__(other)
+
+    def __iter__(self):
+        values = decompose_enum_flags(self)
+        return iter(values)
+
+    def __len__(self):
```
dbyrns, line 349: Is it the number of flags? Are you sure about the implementation?

magpie/models.py
```diff
@@ -132,14 +145,309 @@ def __str__(self):
...
+        users = list(cls.by_status(status=status, db_session=db_session))
+        if not users:
+            return None
+        users = [user for user in users if user.user_name == user_name]
+        if not users:
+            return None
+        if len(users) == 1:
+            return users[0]
+        LOGGER.warning("Duplicate registered/pending users named [%s] detected! This should never happen.", user_name)
+        return users[0] if isinstance(users[0], User) else users[1]
```
dbyrns, line 418: Having more than one user with the same name is indeed suspicious, but having something else at [0]?!? What could be there apart from a `User`?

tests/test_models.py
```diff
@@ -0,0 +1,69 @@
...
+    # this makes it reversed than the above HB -> LB in this case
+    assert list(UserStatuses) == flip_statuses
+    for idx, status in enumerate(reversed(UserStatuses)):
+        assert test_statuses[idx] is status
+
+    idx = None
+
+    # iterate items in random order merged by OR, using all items
+    merge_status = UserStatuses.Pending | UserStatuses.OK | UserStatuses.WebhookError
+    assert len(merge_status) == len(test_statuses)
```
dbyrns, line 47: It looks that the len function is tested :), so it must work.
Will check it again!

dbyrns (approved):
Review completed. Great job!

magpie/models.py
dbyrns: What is the possible values of self.value? Can it be 6 for pending & webhook_error?

magpie/alembic/versions/2021-04-16_dea413e13a8a_pending_users_and_statuses.py
```diff
@@ -0,0 +1,47 @@
...
+    sa.column("status", sa.SmallInteger)
+)
+
+
+def upgrade():
+    op.create_table("users_pending",
+                    sa.Column("id", sa.Integer, primary_key=True, autoincrement=True),
+                    sa.Column("user_name", sa.Unicode(128), unique=True),
+                    sa.Column("user_password", sa.Unicode(256)),
+                    sa.Column("email", sa.Unicode(100), nullable=False, unique=True),
```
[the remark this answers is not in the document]
fmigneault: Doesn't hurt to add the restriction. Still ok for `registered_date` since it is auto-generated.

magpie/api/management/register/register_utils.py
fmigneault: We can do that. Is only a log entry sufficient. 
That info won't be actively stored anywhere, an will be lost according to how frequently the log rotation occurs for example.

magpie/api/management/register/register_utils.py
fmigneault: The page is reused to render different messages based on context, potentially also for other things than user registration (eg T&C "received" indication).
Only needs to provide the value of the message.

magpie/models.py
fmigneault: This code snippet was taken from somewhere in the enum code and related forums. 
There are 2 possible interpretation of `__len__`, depending if the full enum class or a member is processed. 
When full enum, it should return the number of possible distinct flags (3 so far in this case). 
When a member is passed, it must return the number of set flags, which can be a combination of members. 


```python
>>> from magpie.models import UserStatuses
>>> len(UserStatuses)
3
>>> len(UserStatuses.OK)
1
>>> len(UserStatuses.OK|UserStatuses.WebhookError)
2
>>> UserStatuses.WebhookError + UserStatuses.Pending
6
>>> len(UserStatuses.WebhookError|UserStatuses.Pending)
2
```

There are some tests that validate the specific behaviour of `UserStatuses`.

magpie/models.py
fmigneault: This class handles the 2 kind of `User` / `UserPending` entities. 
Technically, using a combination of statutes (`UserStatuses.OK|UserStatuses.Pending` or `"all"`) will run a combined search on the 2 tables, which *could* return distinct entities, with duplicate names. 

I say *technically*, because one would have to manually inject a duplicate name in the other user-type table since each table individually blocks duplicate names with unique constraint. The API also blocks the creation when a duplicate is found in any table. The `[0]` vs `[1]` is in that case is to prioritize the `User` definition over the incomplete `UserPending` one, if that ever happened. Only possible results are `[None, None]`, `[None, User]`, `[None, UserPending]`, `[User, UserPending]` (this edge case) and their swapped around versions since unique are guaranteed from both tables. 

The only way I can think of to guarantee non-duplicate, would be to have a specific `user_name` table with unique constraint, with both `User` and `UserPending` pointing to it by foreign key for their respective names. The only issue about this is that `User` is implemented by ziggurat, and I will need to fiddle with their model. Could work, but could break things unexpectedly somewhere.

magpie/models.py
```diff
@@ -132,14 +145,309 @@ def __str__(self):
...
+            return query.filter((User.user_name == user_name) | (User.email == email)).first()
+        user = cls.by_user_name(user_name=user_name, status=status, db_session=db_session)
+        if user is not None:
+            return user
+        if status is UserStatuses.Pending:
+            return db_session.query(UserPending).filter(UserPending.email == email).first()
+        user = super(UserSearchService, cls).by_email(email=email, db_session=db_session)
+        if user is not None and UserStatuses.get(user.status) in status:
+            return user
+        return db_session.query(UserPending).filter(UserPending.email == email).first()
```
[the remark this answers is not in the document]
fmigneault: Doesn't the docstring provide the necessary details?

magpie/api/management/register/register_utils.py
```diff
@@ -79,3 +139,219 @@ def get_discoverable_group_by_name(group_name, db_session):
...
+    user = pending_user.upgrade(db_session=request.db)
+    LOGGER.debug("Pending user upgraded to full user: [%s (%s)]", user.user_name, user.id)
+
+    # notify the user of its successful account validation, approval and creation
+    params = {"user": user}
+    template = get_email_template("MAGPIE_USER_REGISTRATION_COMPLETED_EMAIL_TEMPLATE", request)
+    ax.evaluate_call(lambda: send_email(user.email, request, template, params),
+                     fallback=lambda: request.db.rollback(), http_error=HTTPInternalServerError,
+                     msg_on_fail="Error occurred during user registration when trying to send "
+                                 "email to pending user for confirmation of its submitted email.")
```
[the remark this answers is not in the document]
fmigneault: I'm using this wording because it refers to the email confirmation step of the full *user registration procedure*, which itself involves many back-and-forth emails. Wording "confirmation of registration" could be confused with confirmation/approval by the admin of that registration.

magpie/cli/send_email.py
```diff
@@ -0,0 +1,112 @@
...
+def make_parser():
+    # type: () -> argparse.ArgumentParser
+    parser = argparse.ArgumentParser(description=__doc__, formatter_class=EmailTemplateChoiceFormatter)
+    email_opts = parser.add_argument_group(title="Email Arguments",
+                                           description="Operation arguments to define email to be sent.")
+    email_opts.add_argument("-e", "--email", required=True, help="Address where to sent the test email.")
+    email_choices = ["test"] + list(DEFAULT_TEMPLATE_MAPPING)
+    email_choices_map = {i: tmpl for i, tmpl in enumerate(email_choices)}
+    email_choices_map.update({str(i): tmpl for i, tmpl in list(email_choices_map.items())})
+    email_choices_map.update({tmpl: tmpl for tmpl in email_choices})
```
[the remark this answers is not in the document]
fmigneault: Indeed comment will help. 
Allows to map `"0" -> "test"` (key as string), `0 -> "test"` (key as int) and literal `"test" -> "test"` to corresponding templates. 
Mostly convenient for the long template names annoying to type, which are easier to select with just the index.
str/int mapping is needed for parsing since int as commandlien arg input becomes a string.

magpie/api/management/register/register_utils.py
fmigneault: I added logs to the sub-operation that handles approve/decline cases.

dbyrns: I don't thing that we must store that info in a db. I only saw a use case where something doesn't run smoothly and seeing your beautiful logs, that was the only missing piece to know the full story.

magpie/models.py
dbyrns: Ok, the loop is pretty genious!

magpie/models.py
dbyrns: Would you mind adding that explanation above the return statement? It makes perfect sense and I don't see any requirements to add a unique username table.

dbyrns: In fact the function main documentation explains that well. I was only trigerred by the warning message just above. Forgot my previous comment.

magpie/models.py
fmigneault: I definitely have no credit. Stack Overflow is your friend :)

dbyrns (approved):
Still good for me

magpie/models.py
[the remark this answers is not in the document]
ChaamC: It does provide a generic overview of the function. Was wondering if we should add more precise details in the different steps of the function too to add some clarity, but if it doesn't add anything, I am ok with leaving it like this too!

tests/test_magpie_ui.py
```diff
@@ -298,6 +298,312 @@ def check_api_resource_permissions(resource_permissions):
...
+                                      msg="Expected email to have been sent to the administrator for approval.")
+                decline_url = wrapped_contents.call_args.args[-1].get("decline_url")
+                decline_path = urlparse(decline_url).path
+                self.login_admin()  # admin needed for decline link
+                resp = utils.test_request(self, "GET", decline_path)
+                body = utils.check_ui_response_basic_info(resp, 200)
+                utils.check_val_is_in("Pending user registration was successfully declined", body)
+                utils.check_val_equal(
+                    mocked_send.call_count, 2,
+                    msg="neither pending user nor notify emails should be sent since registration was declined"
```
ChaamC, line 555: Is there a way for the pending user to know that his registration was declined?

fmigneault: No. It was decided during previous discussions that declines should just be discarded. 
Unless @dbyrns wants to revert that decision such that we also send an email about the decline similarly to the approved email.

ChaamC: I will let you guys take the final decision. But for a user, that has confirmed his registration and is waiting for the admin approval, it could be good to notify him (via email as you suggest) of a request denial instead of leaving him in the unknown.

ChaamC (dismissed):

tests/test_magpie_ui.py
dbyrns: We discussed essentially about the use case where admins would generally approve everyone unless something is suspicious. In that case, decline mail seems indeed superfluous. In a case where we could have a good reason to decline some people (by using a particular domain email for example), it could be helpful for admin to create a decline mail template explaining the guideline under which registration could be refused.

fmigneault: There is no way currently to input some kind of "reason" that the admin could provide to decline the email, just to keep the decline operation simple enough to be a one-button-click operation.
So unless, the template is extremely specific, it will rarely/never cover all cases of potential decline.

dbyrns: I agree. I don't see that as a "this is why I refuse you", but more a "we accept that kind of requests". Maybe the template will not fit the exact case, but at least the user will know that its request as been declined. It also allows to put an email to reach an administrator if wanted.

fmigneault: see latest commits for added decline template email

dbyrns (dismissed):
Perfect!

Issue #433, opened 2021-05-10
[BUG] Intermittent UnauthorizedException for WPS processes behind Magpie 3.8.0 and matching Twitcher


tlvu:
**Describe the bug**
Running Jenkins notebook test suite against Ouranos production PAVICS stack we randomly get 

```
11:22:04  /opt/conda/envs/birdy/lib/python3.7/site-packages/owslib/util.py in openURL(url_base, data, method, cookies, username, password, timeout, headers, verify, cert, auth)
11:22:04      210     if req.status_code in [400, 401]:
11:22:04  --> 211         raise ServiceException(req.text)
11:22:04      212 
11:22:04  
11:22:04  ServiceException: <?xml version="1.0" encoding="utf-8"?>
11:22:04  <ExceptionReport version="1.0.0"
11:22:04      xmlns="http://www.opengis.net/ows/1.1"
11:22:04      xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
11:22:04      xsi:schemaLocation="http://www.opengis.net/ows/1.1 http://schemas.opengis.net/ows/1.1.0/owsExceptionReport.xsd">
11:22:04      <Exception exceptionCode="NoApplicableCode" locator="AccessForbidden">
11:22:04          <ExceptionText>Not authorized to access this resource. User does not meet required permissions.</ExceptionText>
11:22:04      </Exception>
11:22:04  </ExceptionReport>
11:22:04  
11:22:04  During handling of the above exception, another exception occurred:
11:22:04  
11:22:04  UnauthorizedException                     Traceback (most recent call last)
11:22:04  <ipython-input-5-a503678973ad> in <module>
11:22:04       30     elevation=basin_elevation,
11:22:04       31     area=basin_area,
11:22:04  ---> 32     name="test_basin",
11:22:04       33 )
11:22:04       34 
11:22:04  
11:22:04  </opt/conda/envs/birdy/lib/python3.7/site-packages/birdy/client/base.py-613> in climatology_esp(self, forecast_date, ts, model_name, params, forecast_duration, name, area, latitude, longitude, elevation, nc_spec, rvc, rain_snow_fraction, output_formats)
11:22:04  
11:22:04  /opt/conda/envs/birdy/lib/python3.7/site-packages/birdy/client/base.py in _execute(self, pid, **kwargs)
11:22:04      357             if "AccessForbidden" in str(e):
11:22:04      358                 raise UnauthorizedException(
[...]

PR #439, opened 2021-07-10
Resolve permission error resolution related to service cache with OWS parser

fmigneault:
# Changes

## Fix caching issue
 Fix an issue in ``MagpieAdapter`` when `Service` caching is enabled (in `Twitcher` INI configuration) that caused
  implementations derived from ``ServiceOWS`` (WPS, WMS, WFS) to incorrectly retrieve and parse the cached request
  parameters instead of the new ones from the incoming request.
  Because ``ServiceOWS`` implementations employ request parameter ``request`` (in query or body based on HTTP method)
  to infer their corresponding `Magpie` ``Permission`` (e.g.: ``GetCapabilities``, ``GetMap``, etc.), this produced
  potential inconsistencies between the *requested* ``Permission`` that `Twitcher` was evaluating with `Magpie`, and
  the *actual request* sent to the `Service` behind the proxy.

## More detailed response

This revolves around the error message indicated in #433 that initiated my investigation, which lead to finding  the above cache fix.
The error message was not wrong though, just slightly misleading. 
When OWS service cannot resolve the requested permission (in Twitcher via `MagpieAdapter`), the message will be more precise about it.

### Old response
code: 500 and `NoApplicableCode` (aka `NotImplementedError` raised) returned regardless of the query parameters

image

### New response
code: 400 and either `InvalidParameterValue` or `MissingParameterValue` returned appropriately. 
plus, `locator` is provided as well.

image


magpie/services.py
```diff
@@ -412,21 +417,40 @@ class ServiceOWS(ServiceInterface):
...
+        return self._request
+
+    def _set_request(self, request):
+        # type: (Request) -> None
+        self._request = request
+        # must reset the parser from scratch if request changes to ensure everything is updated with new inputs
         self.parser = ows_parser_factory(request)
         self.parser.parse(self.params_expected)  # run parsing to obtain guaranteed lowered-name parameters
 
+    request = property(_get_request, _set_request)
```
dbyrns, line 435: So, does this new property fix this issue https://github.com/Ouranosinc/Magpie/issues/433 ?
It looks like this is the only thing not related to improve exception reporting...

dbyrns (dismissed):
It looks good. I was wondering on `self.get_services(request)`, but it looks like Twitcher can properly raise a `ServiceNotFound` (https://github.com/bird-house/twitcher/blob/master/twitcher/utils.py#L78) if the service name cannot be retrieved from the url and if the service name is not known we should get a `HTTPNotFound` (https://github.com/Ouranosinc/Magpie/blob/master/magpie/adapter/magpieowssecurity.py#L67). 

So the stress test is now running properly?

magpie/services.py
fmigneault: I've relaunched the stress test on a daccs-iac instance with [bird-house/birdhouse-deploy@cache-settings]( https://github.com/bird-house/birdhouse-deploy/tree/cache-settings) and still see the 401. I cannot replicate though when only using `master` without cache. I'm not sure yet if they are related or not. Caching only seems to make it easier to catch the 401 since it will repeat over multiple requests rather than one-off call.

dbyrns: With the new exception information, do we have a better clue on what is happening? What is the request params and why Magpie doesn't find the permission with it?

fmigneault: I think I might have found the cause. 

When service is retrieved from the cache in MagpieAdapter, and gets forward to `effective_permissions` (method that resolves access with ACL), it tries to use that service. 
There is no database session associated to that service (since from cache), so somewhere deep in ziggurat_foundation code, `session.query` with `session=None` gets called, and raises `AttributeError`. 
Because of all the try/except, Twitcher simply catches it as returns "security check failed", can because the user happen to be anonymous, it indicates 401. 
What is tricky is that when the ACL itself was cached, this check was bypassed, which is why 200 was returned as normal. 
Same goes for the first request since service is not cached, so `effective_permisisons` are resolved with a valid session. 

Really a confusion time-specific situation!

tlvu (dismissed):
+1, I reviewed the very well explained PR description :D  Always a good idea to have a clear error message to help diagnose problems.  I trust you guys with the code change.

magpie/services.py
fmigneault: Should be adjusted with db-reconnect to service instance (see 0f99bdad0dc97762c97657c42f61b6f0b5e698d0)

dbyrns (approved):
Great job on the tests!

PR #461, opened 2021-08-10
fix and improve handling of parameter update of service API request

fmigneault:
## changes 

* Improve API update operation of ``Service`` for allowed fields in order to accept body containing only the
  new value for the custom ``configuration`` without additional parameters. It was not possible to
  update ``configuration`` by itself, as ``service_name`` and ``service_url`` were independently validated
  for new values beforehand.
* Fix erasure value ``None`` (JSON ``null``) validation when updating ``Service`` field ``configuration`` to
  properly distinguish explicitly provided ``None`` against default value when the field is omitted.
* Fix incorrect OpenAPI body schema indicated in response of ``POST /services`` request.
* Updated tests for new use cases.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,12 +9,18 @@ Changes
 
 Features / Changes
 ~~~~~~~~~~~~~~~~~~~~~
-* N/A
+* Improve API update operation of ``Service`` for allowed fields in order to accept body containing only the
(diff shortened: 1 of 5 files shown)


dbyrns:
LGTM. Don't hesitate to ask someone else in the team to do the review if I cannot do it in a timely manner.

Issue #335, opened 2020-05-15
[Feature] Add tooltip details to some items


fmigneault:
# Context

Some items are sometimes hard to guess the functionality from a user that did not directly participate in detailed implementation of some components. 

A few examples to address (not exhaustive, add/edit as needed): 
- [ ] distinction between `-match` and "regular" corresponding permissions is not obvious at all
- [ ] with #334, distinction between `read`/`write` and request method specific permissions
- [ ] ... ?

# Feature

The feature consists of simply adding `tooltip` HTML/CSS elements that would help provide a brief description of the corresponding item. 
A small hover/popup-like style (maybe with a (i) icon also) can be used so that the UI doesn't get too cluttered.
example doc: https://www.w3schools.com/css/css_tooltip.asp

PR #467, opened 2021-09-21
Feature add tc

ChaamC:
Features / Changes
------------------------------
* Add new ``Terms and conditions`` field for ``group`` creation. When a request is made to assign a ``user`` to a
  ``group`` with terms and conditions, an email is now sent to the ``user`` with the terms and conditions. The ``user``
  is assigned to the ``group`` when receiving the ``user``'s approval of terms and conditions, and another email is
  then sent to notify the ``user`` of the successful operation.
* Add new requests under ``/groups/{group_name}/pending_users``, ``/users/current/pending_groups`` and
  ``/users/{user_name}/pending_groups`` endpoints to get pending ``users`` or ``groups``. These routes are used to
  display any pending ``users``/``groups`` on the UI.

Bug Fixes
--------------------------
* Fix HTTP ``Internal Server Error [500]`` on the page to edit a group when deleting the last ``user`` of a ``group``.

Other references
----------------------
* Implements feature [DAC-119](https://www.crim.ca/jira/browse/DAC-119)


magpie/api/management/user/user_utils.py
```diff
@@ -311,6 +311,89 @@ def assign_user_group(user, group, db_session):
...
+    ax.evaluate_call(lambda: send_email(user.email, settings, template, params),
+                     fallback=lambda: db_session.rollback(), http_error=HTTPInternalServerError,
+                     msg_on_fail="Error occurred while adding user to a group when trying to send "
+                                 "email to user for requesting agreement of the group's terms and conditions.")
+
+    return ax.valid_http(http_success=HTTPAccepted, detail=s.UserGroups_POST_AcceptedResponseSchema.description,
+                         content={"user_name": user.user_name, "group_name": group.group_name})
+
+
+def request_assign_user_group(user, group, db_session):
```
fmigneault, line 344: The name should be more obvious about the conditional aspect. 
Something like `create_pending_or_assign_user_group` ? 

Prefix `request` is ambiguous here since HTTP request seems to be already handled prior to this call (user/group passed directly and pre-validated).

magpie/api/management/user/user_utils.py
```diff
@@ -311,6 +311,89 @@ def assign_user_group(user, group, db_session):
...
+    # type: (models.User, models.Group, Session) -> None
+    """
+    Requests the creation of a user-group relationship (user membership to a group).
```
fmigneault, line 347: I would put more emphasis on the conditional aspect in the first sentence.
Something like `"Creates a user-group ... or pending T&C confirmation."`

magpie/api/management/user/user_utils.py
```diff
@@ -311,6 +311,89 @@ def assign_user_group(user, group, db_session):
...
+
+def handle_user_group_terms_confirmation(tmp_token, request):
+    # type: (models.TemporaryToken, Request) -> Response
+    """
+    Handles the confirmation of a user to accept the terms and conditions of a group.
+
+    Generates the appropriate response that will be displayed to the user.
+    """
+    LOGGER.info("User %s approved terms and conditions of group %s.",
+                tmp_token.user.user_name, tmp_token.group.group_name)
```
fmigneault, line 370: Please also provide the user and group IDs here.
Place the `%s` fields within `[%s]` to make it more obvious in logs about the start/end of values.

magpie/api/management/user/user_utils.py
```diff
@@ -311,6 +311,89 @@ def assign_user_group(user, group, db_session):
...
+    template = get_email_template("MAGPIE_GROUP_TERMS_APPROVED_EMAIL_TEMPLATE", request)
+    ax.evaluate_call(lambda: send_email(tmp_token.user.email, request, template, params),
+                     fallback=lambda: request.db.rollback(), http_error=HTTPInternalServerError,
+                     msg_on_fail="Error occurred during group terms confirmation when trying to send "
+                                 "email to user for confirmation of the terms and conditions acceptation.")
+
+    # Remove all group_accept_terms temporary tokens associated with the same user and group
+    tmp_tokens = TemporaryToken.by_user(tmp_token.user)\
+        .filter(TemporaryToken.operation == TokenOperation.GROUP_ACCEPT_TERMS)\
+        .filter(TemporaryToken.group == tmp_token.group)
```
fmigneault, line 384: Avoid `\` using reassign such as: 
```python 
tmp_tokens = TemporaryToken.by_user(tmp_token.user)
tmp_tokens = tmp_tokens.filter(TemporaryToken.operation == TokenOperation.GROUP_ACCEPT_TERMS)
tmp_tokens = tmp_tokens.filter(TemporaryToken.group == tmp_token.group)
```

magpie/api/notifications.py
```diff
@@ -23,6 +23,10 @@
 
 TEMPLATE_DIR = os.path.join(os.path.dirname(os.path.abspath(__file__)), "templates")
 DEFAULT_TEMPLATE_MAPPING = {
+    "MAGPIE_GROUP_TERMS_APPROVED_EMAIL_TEMPLATE":
+        os.path.join(TEMPLATE_DIR, "email_group_terms_approved.mako"),
+    "MAGPIE_GROUP_TERMS_SUBMISSION_EMAIL_TEMPLATE":
+        os.path.join(TEMPLATE_DIR, "email_group_terms_submission.mako"),
```
fmigneault, line 29: These need to be added to docs since they can be overridden by configuration. 
Probably a new section in https://raw.githubusercontent.com/Ouranosinc/Magpie/master/docs/configuration.rst under the `User Registration and Approval Configuration` section.
Must also be added to below `get_email_template` function.

magpie/api/templates/email_group_terms_submission.mako
```diff
@@ -0,0 +1,56 @@
...
+
+        <p>
+        A request to ${magpie_url} has been submitted to join the '${group_name}' group.
+        Before you can be assigned to the group, we need your consent to the group's terms and conditions.
+        </p>
+
+        <p style="font-weight: bold;">
+            Terms and conditions:
+        </p>
+        <p style="margin-left: 40px; font-style: italic;">
```
fmigneault, line 45: nice

magpie/ui/management/templates/add_group.mako
```diff
@@ -71,6 +71,31 @@
             <td>
             </td>
         </tr>
+        <tr>
+            <td>Terms and conditions:</td>
+            <td>
+                <label>
+                    <input type="text" name="terms" value="${form_terms}" />
```
fmigneault, line 78: Hard to know with static analysis. 
Is this field sufficiently large for convivial writing of long blocks of T&C text? 
I think the field is by default not expandable and single line. 

Might need to increase its size for this case + ensure text wrapping is used to help writing paragraphs instead of horizontally scrolling text box.
You can make a specific CSS class to apply needed styles to apply this behaviour if other similar cases arise.

magpie/ui/management/templates/edit_user.mako
```diff
@@ -230,7 +230,13 @@
...
                 %endif
                >
-            ${group}
+            %if group in pending_groups:
+                <!-- checkbox is not checked or disabled for pending groups
+                     so additional requests and emails can still be sent if needed -->
+                ${group} [pending]
+            %else:
+                ${group}
+            %endif
```
fmigneault, line 239: Self note: add tooltip for this #335

magpie/ui/management/templates/edit_group.mako
```diff
@@ -144,34 +144,53 @@
...
+                %endif
+            >
+            %if user in pending_users:
+                <!-- checkbox is not checked or disabled for pending users
+                     so additional requests and emails can still be sent if needed -->
+                ${user} [pending]
             %else:
-               onchange="document.getElementById('edit_members').submit()"
+                ${user}
             %endif
```
fmigneault, line 188: Self note: add tooltip for this #335

magpie/ui/user/templates/edit_current_user.mako
```diff
@@ -242,7 +242,13 @@
...
                             %endif
                         >
-                        ${group}
+                        %if group in pending_groups:
+                            <!-- checkbox is not checked or disabled for pending groups
+                                 so additional requests and emails can still be sent if needed -->
+                            ${group} [pending]
+                        %else:
+                            ${group}
+                        %endif
```
fmigneault, line 251: Self note: add tooltip for this #335

fmigneault (changes requested):
Very nice PR and great tests! 
I would like only a modification regarding the API endpoints (see comment in CHANGES). 
Need a missing section in docs. 
Code of the feature look all good.

magpie/api/management/group/group_views.py
```diff
@@ -31,7 +39,8 @@ def create_group_view(request):
     group_name = ar.get_value_multiformat_body_checked(request, "group_name")
     group_desc = ar.get_multiformat_body(request, "description", default="")
     group_disc = asbool(ar.get_multiformat_body(request, "discoverable", default=False))
-    return gu.create_group(group_name, group_desc, group_disc, request.db)
+    group_terms = ar.get_multiformat_body(request, "terms", default="")
```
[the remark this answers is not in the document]
ChaamC: Good point. For consistency, should the `description` field have a default value of `None`, since it is also a nullable string?

magpie/ui/management/templates/add_group.mako
ChaamC: Good idea, I didn't think to test for long blocks of text, I will look into it!

magpie/api/notifications.py
ChaamC: I'll add template variables in the magpie.ini also, like you did for the user registration templates.

CHANGES.rst
```diff
@@ -7,8 +7,19 @@ Changes
 `Unreleased <https://github.com/Ouranosinc/Magpie/tree/master>`_ (latest)
 ------------------------------------------------------------------------------------
 
+Features / Changes
+~~~~~~~~~~~~~~~~~~~~~
+* Add new ``Terms and conditions`` field for ``group`` creation. When a request is made to assign a ``user`` to a
+  ``group`` with terms and conditions, an email is now sent to the ``user`` with the terms and conditions. The ``user``
+  is assigned to the ``group`` when receiving the ``user``'s approval of terms and conditions, and another email is
+  then sent to notify the ``user`` of the successful operation.
+* Add new requests under ``/groups/{group_name}/pending_users``, ``/users/current/pending_groups`` and
```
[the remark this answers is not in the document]
ChaamC: So if I understand, for example, I can move the code of `/groups/{group_name}/pending_users` to `/groups/{group_name}/users`, and add a new input parameter to the `/groups/{group_name}/users` query, indicating what type of users to get, with an int.
For example :
`ACTIVE_USERS = 0`
`PENDING_USERS = 1`
`ALL_USERS = 2` (which adds to each returned user the boolean `pending` field you mentioned)

I can add a `user_type` parameter to the input of the GET query, which contains an int corresponding to one of the entry above.

magpie/ui/management/templates/add_group.mako
ChaamC: I am currently checking to make it better for multiline text. I use a `textarea`, that can be resized only vertically. By default, it displays 4 rows, but can be resized to display more, and can be scrolled through too. What do you think of the layout? 

This is just a fast draft, but I think I should make the `Terms and conditions` and its associated `(optional)` label aligned with the top of the `textarea`. 

Also, I should align the `(optional)` label of the `description` beside its input text box, and not aligned with the `textarea` of the T&C. 

There is the `Add Group` button that should stay aligned as before, with the input text boxes of `Group name` and `Description`, or it will move depending of the horizontal size of the `textarea`. You can see it has a weird alignment on the current screenshot.

Also, what do you think of the horizontal size of the `textarea`? I didn't want to make the same size as the others, as it should contain more text and should be larger. I think I should leave the horizontal resizing enabled too, so the user can use the width he wants.

Let me know if my idea makes sense or if you have any better suggestion so I can work on a proper version of the ui!

image

ChaamC: I've added a minimal width and height, to avoid weird formatting, and realigned the different elements.

Here is how it looks, by default, when opening the page :
image

Here is an example with the minimal size (the default width is a bit larger) :
image

Here is an example with the expanded textarea (the different elements mentioned before are now aligned):
image

CHANGES.rst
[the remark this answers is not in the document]
fmigneault: Yes. Exactly what I add in mind. You should only need to remap the detected `"pending"` to your function that was defined in current `pending_users` endpoint, and if `"all"` was requested, extend the result with the usual operation.

magpie/ui/management/templates/add_group.mako
fmigneault: > I use a textarea, that can be resized only vertically. By default, it displays 4 rows, but can be resized to display more, and can be scrolled through too. What do you think of the layout?

Excellent!

> This is just a fast draft, but I think I should make the Terms and conditions and its associated (optional) label aligned with the top of the textarea.

I totally agree.

> Also, I should align the (optional) label of the description beside its input text box, and not aligned with the textarea of the T&C.
> Also, what do you think of the horizontal size of the textarea? I didn't want to make the same size as the others, as it should contain more text and should be larger. I think I should leave the horizontal resizing enabled too, so the user can use the width he wants.

The last screenshot looks perfect. 
If `(optional)` works by automatically following resizes of T&C, let's keep that design.

ChaamC:
@fmigneault Branch updated according to your feedback, let me know if you have any comment on this new update!

Also, for the version bump, I saw that there was already some changes in the `Unreleased` section of the `CHANGES.rst` file. When the PR is accepted, do I just merge my branch, and you will do a bump eventually, or should I make a bump version to 3.16.0 before merging?

magpie/api/management/user/user_views.py
```diff
@@ -124,35 +124,26 @@ def delete_user_view(request):
...
-                                  api_security=s.SecurityAuthenticatedAPI)
-@view_config(route_name=s.UserPendingGroupsAPI.name, request_method="GET", permission=MAGPIE_CONTEXT_PERMISSION)
-def get_user_pending_groups_view(request):
-    """
-    List all groups where the user's membership is pending, because of required terms and conditions acceptation.
-    """
-    user = ar.get_user_matchdict_checked_or_logged(request)
-    tmp_tokens = TemporaryToken.by_user(user).filter(TemporaryToken.operation == TokenOperation.GROUP_ACCEPT_TERMS)
-    pending_group_names = [tmp_token.group.group_name for tmp_token in tmp_tokens]
+    group_type = ar.get_multiformat_body(request, "group_type", default=UserGroupType.ACTIVE_USERGROUPS.value)
```
fmigneault, line 131: Requires validation of the value. 
This also doesn't fetch from the query string, but rather the request body (not available for GET).

I think tests are in order also to ensure these cases: 
- old/default listing is returned if nothing/"active" is provided
- only pending groups are returned when asking for it
- both are returned when "all"

magpie/models.py
```diff
@@ -101,6 +101,15 @@ class GroupPermission(GroupPermissionMixin, Base):
     pass
 
 
+class UserGroupType(ExtendedEnum):
+    """
+    Supported types of user-group relationships.
+    """
+    ALL_USERGROUPS = "all_usergroups"
+    ACTIVE_USERGROUPS = "active_usergroups"
+    PENDING_USERGROUPS = "pending_usergroups"
```
fmigneault, line 110: Remove redundant parts `_USERGROUPS` from names. 
Only `ALL = "all"`, etc. are sufficient. 
This will make requests easier to call/remember also:
```http
GET <>/users/me/groups?status=pending
GET <>/groups/people/users?status=all
```

magpie/api/management/group/group_views.py
```diff
@@ -124,48 +117,31 @@ def delete_group_view(request):
...
-    Pending users are users who requested to join a group but still need to accept terms and conditions.
-    """
-    # Find the group associated with the request
-    group_name = ar.get_value_matchdict_checked(request, "group_name")
-    group = ax.evaluate_call(lambda: GroupService.by_group_name(group_name, db_session=request.db),
-                             fallback=lambda: request.db.rollback(), http_error=HTTPForbidden,
-                             msg_on_fail=s.GroupPendingUsers_GET_ForbiddenResponseSchema.description)
-    ax.verify_param(group, not_none=True, http_error=HTTPNotFound,
-                    msg_on_fail=s.Group_MatchDictCheck_NotFoundResponseSchema.description)
+    user_type = ar.get_multiformat_body(request, "user_type", default=UserGroupType.ACTIVE_USERGROUPS.value)
```
fmigneault, line 124: must validate value and fetch from query

magpie/api/management/group/group_views.py
```diff
@@ -124,48 +117,31 @@ def delete_group_view(request):
...
-
-    # Find and return all user names associated with the discovered tokens
-    pending_user_names = [tmp_token.user.user_name for tmp_token in tmp_tokens]
-
-    # Remove any user already belonging to the group, in case any tokens are irrelevant.
-    # Should not happen since related tokens are deleted upon T&C acceptation.
+    user_names = []
     member_user_names = ax.evaluate_call(lambda: [user.user_name for user in group.users],
                                          http_error=HTTPForbidden,
                                          msg_on_fail=s.GroupUsers_GET_ForbiddenResponseSchema.description)
```
fmigneault: skip unnecessary fetch requested is pending only

magpie/api/management/group/group_views.py
```diff
@@ -124,48 +117,31 @@ def delete_group_view(request):
...
+        # Find all temporary tokens with requested group id that have a pending accept terms request
+        tmp_tokens = request.db.query(TemporaryToken).filter(TemporaryToken.group_id == group.id)
+        tmp_tokens = tmp_tokens.filter(TemporaryToken.operation == TokenOperation.GROUP_ACCEPT_TERMS)
+
+        # Find and return all user names associated with the discovered tokens
+        pending_user_names = [tmp_token.user.user_name for tmp_token in tmp_tokens]
+
+        # Remove any user already belonging to the group, in case any tokens are irrelevant.
+        # Should not happen since related tokens are deleted upon T&C acceptation.
+        user_names += [usr for usr in pending_user_names if usr not in member_user_names]
```
fmigneault, line 142: Uou can take advantage of sets to do this faster.
Define initial `user_names` as `set()` and `union` with anything to add. 
Then convert as back to list with `sorted(list(user_names))` call below.

magpie/api/management/user/user_views.py
```diff
@@ -124,35 +124,26 @@ def delete_user_view(request):
...
+        pending_group_names = [tmp_token.group.group_name for tmp_token in tmp_tokens]
+
+        # Remove any group a user already belongs to, in case any tokens are irrelevant.
+        # Should not happen since related tokens are deleted upon T&C acceptation.
+        group_names += [grp for grp in pending_group_names if grp not in member_group_names]
 
-    return ax.valid_http(http_success=HTTPOk, content={"pending_group_names": pending_group_names},
-                         detail=s.UserPendingGroups_GET_OkResponseSchema.description)
+    return ax.valid_http(http_success=HTTPOk, content={"group_names": group_names},
+                         detail=s.UserGroups_GET_OkResponseSchema.description)
```
fmigneault: same comments as for group_users endpoint case

magpie/ui/home/static/style.css
```diff
@@ -1268,3 +1268,9 @@ div.input-container {
     height: 100%;
     -webkit-box-sizing: border-box;
 }
+
+textarea {
+    resize: both;
+    font-size: 13.3333px;
+    font-family: arial, sans-serif;
```
fmigneault, line 1275: Are the font parts needed? 
I think there are generic ones applied for the whole CSS.

It it is indeed necessary, you can remove them from here and add `textarea` at the same place where other fonts are defined.

magpie/ui/management/templates/add_group.mako
```diff
@@ -72,13 +73,13 @@
...
-            <td>
+            <td style="vertical-align: top">Terms and conditions:</td>
+            <td colspan="2">
                 <label>
-                    <input type="text" name="terms" value="${form_terms}" />
+                    <textarea rows="4" cols="50" style="min-width: 282px; min-height: 66px" name="terms">${form_terms}</textarea>
                 </label>
             </td>
-            <td>
+            <td style="vertical-align: top">
```
fmigneault: Use a specific CSS class for `vertical-align` parts.
Use another CSS class (eg: `field-expandable`) to define anything specific to that feature.
There should be other field/table classes in the style file that affect the display of elements of this part. 

Employ `%` or `em` rather than `px` since they are different for each monitor. 
They seem specific for your display but could appear wrong for others.

fmigneault (changes requested):
Schema `UserGroups_GET_RequestSchema` and `GroupUsers_GET_RequestSchema` are missing the new query parameter. 
It should not use `body` because it is really inconvenient to do browser GET requests with a body, and some newer HTTP recommendations even say to drop the body in GET. 

I think it should also be renamed simply `status` so you can reuse the same definition for both endpoints, since they basically correspond to the same thing. The enum can be adjusted to be `UserGroupStatus`. 

Look for `querystring = ...` in another schema for and example how to define the  query part of requests. 
The parameter should use `colander.OneOf` with the `UserGroupType` enum for allowed values.

dbyrns (commented):
I made a quick check to review the use case coverage and everything looks good. I'm also happy with the extensive tests.
If possible, I would like to access a live instance of this version to validate the UI too.

magpie/api/management/user/user_views.py
ChaamC: If that works for you, I will complete those 2 tests ([test_GetUserGroups](https://github.com/Ouranosinc/Magpie/blob/9deeda69e50bbc4f06ce1f3670f09a5bfc6cafae/tests/interfaces.py#L4092), [test_GetGroupUsers](https://github.com/Ouranosinc/Magpie/blob/9deeda69e50bbc4f06ce1f3670f09a5bfc6cafae/tests/interfaces.py#L4475)) by adding all the possible cases (all, active, pending) of the GET queries.

fmigneault: I would rather have separate tests that have explicit queries so the current ones remain simple.

ChaamC: I am not sure I understand. By separate tests, you mean one test for each case? 
So basically :
4 different test cases (all, pending, active, and no specified status)
x 2 query types (for GetUserGroups and GetGroupUsers)
So 8 tests in total.
The 2 existing ones I linked in my last comment would correspond to 2 of those 8 cases.

I think it would make more sense to keep all 4 different test cases in one single test for each query type. So, we keep just the 2 original tests.
This way, I can setup the user/groups just one time. For example, for the GetUserGroups query, I just have to add the user to a group with terms, and to a group without terms, and then check with the 4 GET query status argument cases, if the results are as expected.

magpie/api/management/group/group_views.py
ChaamC: I currently use the member_user_names variable even for pending only.
I could remove it if it is too much, but I was removing any member names found in the temporary tokens, in case any tokens was remaining for an existing member. It is not supposed to happen, but if any token would remain, it would indicate the user as pending, even if he was really a member, since we rely on the tokens to find pending users.

Technically, I do remove any tokens, when the T&C are accepted, but I was doing this in case.
Let me know if you'd rather I remove this step.

magpie/ui/home/static/style.css
ChaamC: It is necessary as you can see on this image, if I remove those font parts : 

I checked and if not defined, it uses textarea's user agent stylesheet, which uses `monospace` font-family instead of `arial`.
I put those font settings to reproduce the settings used in the other `input[type="text"]` found on the same page, but by looking at those other text inputs, I see that they also use the default user agent stylesheet. It's just that the default user agent stylesheet settings are different between `input[type="text"]` and `textarea`. So, I will keep those font settings in the file, but wondering if it is okay to leave the `input[type="text"]` with the default settings. Let me know if I should adjust anything!

Also, not really sure, I don't really see a specific place where the other fonts are defined, there are a few definitions scattered in the file, but not really in one specific place. I think I will just move it to the more generic section near the top of the file.

magpie/api/management/user/user_views.py
ChaamC: I pushed the 2 completed tests, which each include the 4 status test cases, let me know what you think.

ChaamC:
> I made a quick check to review the use case coverage and everything looks good. I'm also happy with the extensive tests. If possible, I would like to access a live instance of this version to validate the UI too.

@dbyrns Thanks for the feedback. For the UI validation, I don't think we have an easy way to test it, like an accessible test instance or something. You can start a local instance on your machine though, just by starting the docker-compose found on the repo (need to set HOST_FQDN variable too, `export HOST_FQDN=localhost` for example), and accessing `localhost:2001`

magpie/api/management/user/user_views.py
fmigneault: With separate tests it is easier to identify if only one case fails for some underlying reason or all of them are invalid. 
When combined in one test, the first error fails the test, but we don't get the details of (possible) other failing cases.

You can make an "util" method that does the same procedure with an input query and expected members to reduce duplication of code for all combinations.

magpie/api/management/group/group_views.py
fmigneault: Ok. I see. Then it's fine to leave it like this.

magpie/ui/home/static/style.css
fmigneault: It's ok to apply the font specifically for `textarea`. 
I would like to have one place where they are all defined. 

So somewhere at the top under the `html` and `body` part, you can define this: 

```css
table,
textarea,
input[type="text"],
.panel-line    /* others??? */ {  
    font-family: arial, sans-serif;
	font-size: inherit;    /* needed? */
}
```

I'm still not sure about the `font-size: 13.3333px;` though. 
It is hard to tell who everything resolves by static analysis will the overrides that occurs in there. 

It should use the same value as the text inputs. 
If its not the case, you can also apply it to the same 'font' style.

dbyrns:
> @dbyrns Thanks for the feedback. For the UI validation, I don't think we have an easy way to test it, like an accessible test instance or something. You can start a local instance on your machine though, just by starting the docker-compose found on the repo (need to set HOST_FQDN variable too, `export HOST_FQDN=localhost` for example), and accessing `localhost:2001`

I launch my own instance and failed to check the full workflow. First, I did not configure SMTP, and trying to check a t&c group in the admin user dashboard I got an internal server error with full stack trace. (@fmigneault, not sure if this is something we really want). I then proceed to properly configure the SMTP server (I double check with smtplib to be sure that the parameters are working). Using the same scenario, I got a [pending] beside my user, but because no email has been send I check the log, only to discover a SMTPSenderRefused error. So : 

1. If the t&c mail cannot be sent properly, we should replace the [pending] by an error status
2. I check the notification doc at (https://pavics-magpie.readthedocs.io/en/stable/_modules/magpie/api/notifications.html) to be sure that I set the good SMTP parameters and even succeed to send a mail directly with smtplib using those. I wonder why I cannot configure properly Magpie?

ChaamC:
> I launch my own instance and failed to check the full workflow. First, I did not configure SMTP, and trying to check a t&c group in the admin user dashboard I got an internal server error with full stack trace. (@fmigneault, not sure if this is something we really want). I then proceed to properly configure the SMTP server (I double check with smtplib to be sure that the parameters are working). Using the same scenario, I got a [pending] beside my user, but because no email has been send I check the log, only to discover a SMTPSenderRefused error. So :
> 
> 1. If the t&c mail cannot be sent properly, we should replace the [pending] by an error status
> 2. I check the notification doc at (https://pavics-magpie.readthedocs.io/en/stable/_modules/magpie/api/notifications.html) to be sure that I set the good SMTP parameters and even succeed to send a mail directly with smtplib using those. I wonder why I cannot configure properly Magpie?

@dbyrns 

1. It makes sense that the user should not be pending if an error occured while sending email. I guess, as you say, it should display an error occured, and should reset the user as not being a member of the group. I will look into it.
2. To test sending emails, I used 3 variables in the `magpie.ini` file : `magpie.smtp_from` (a working email address), `magpie.smtp_host` (I used value `smtp.gmail.com` as my email address was gmail) and `magpie.smtp_password` (the password of the email adress used for first variable.) I don't know if that helps?

dbyrns:
Thank you @ChaamC, using the ini rather than the env variables did the trick. @fmigneault , I'm still not sure why the ini has priority over env variables since the ini is embedded in the image while env variables can easily be overriden, but I guess that you face problems the other way around too. 

As for the UX review, I really like how everything work. Good job! However I think that the two following things should be improved : 
1. As said previously, any error that prevent the t&c mail to be sent should not yield a [pending] label beside the username, but maybe show an error banner?
2. In the self-serve account page, when a user self-register to a group having t&c, the [pending] label also appears without any other indication. I think that a banner should also appear saying that a mail has been send with the t&c and must be processed to complete the group registration.

fmigneault:
> Thank you @ChaamC, using the ini rather than the env variables did the trick. @fmigneault , I'm still not sure why the ini has priority over env variables since the ini is embedded in the image while env variables can easily be overriden, but I guess that you face problems the other way around too.

This is a design choice. It is described here (top most): https://pavics-magpie.readthedocs.io/en/latest/configuration.html
ENV will work if you don't have a conflict. I try to move away from ENV as must as possible since it is always possible to modify them while the app is running. That could break some parts of the code that assume values don't change midway. INI are loaded and constant (unless explicitly overridden in code of course).

fmigneault:
>     1. It makes sense that the user should not be pending if an error occured while sending email. I guess, as you say, it should display an error occured, and should reset the user as not being a member of the group. I will look into it.

>     1. As said previously, any error that prevent the t&c mail to be sent should not yield a [pending] label beside the username, but maybe show an error banner?

The `Users.status` filed could be used for that. 
Another state could be added to `UserStatuses` (`OK`, `WebhookError`, `Pending` at the moment).

fmigneault:
> I launch my own instance and failed to check the full workflow. First, I did not configure SMTP, and trying to check a t&c group in the admin user dashboard I got an internal server error with full stack trace. (@fmigneault, not sure if this is something we really want). I then proceed to properly configure the SMTP server (I double check with smtplib to be sure that the parameters are working). Using the same scenario, I got a [pending] beside my user, but because no email has been send I check the log, only to discover a SMTPSenderRefused error. So :

@dbyrns You can use the CLI utility `magpie_cli send_email` to test if you SMTP config is valid. It is quicker than doing the full User T&C loop.

fmigneault:
@ChaamC 
Will need to resolve conflicts in readme. 
We might need to synchronize also regarding this PR, #470 and #471.

dbyrns:
> > I launch my own instance and failed to check the full workflow. First, I did not configure SMTP, and trying to check a t&c group in the admin user dashboard I got an internal server error with full stack trace. (@fmigneault, not sure if this is something we really want). I then proceed to properly configure the SMTP server (I double check with smtplib to be sure that the parameters are working). Using the same scenario, I got a [pending] beside my user, but because no email has been send I check the log, only to discover a SMTPSenderRefused error. So :
> 
> @dbyrns You can use the CLI utility `magpie_cli send_email` to test if you SMTP config is valid. It is quicker than doing the full User T&C loop.

Thank you, but once I remembered the ini precedence over env it has been easy to configure. As for the reason why this precedence is used, I agree that in production it makes sense. Just a little bit more involving for casual dev debugging 😜

ChaamC:
> The `Users.status` field could be used for that. Another state could be added to `UserStatuses` (`OK`, `WebhookError`, `Pending` at the moment).

@fmigneault I am not sure I understand the use of a new UserStatus here. 
In the case of a fail, when trying to assign a user with a group with T&C, I would just reset the page, making sure not to display a `pending` label for the requested group, removing any related TemporaryTokens, and displaying an error message on the ui.

The error could come from different causes. For example, an invalid smtp configuration could cause an error, and it is not necessarily something related to the user here, so I don't know if we should keep the error state on the user's status.

ChaamC:
Updated with latest requested changes.

@fmigneault 
@dbyrns I wasn't able to reproduce the behaviour you had exactly, having the "[pending]" displayed when the email failed, but maybe the latest changes fixed it? I tested the ui with empty smtp variables values in magpie.ini, and it just resets the page, membership staying inactive. Let me know if you are still able to reproduce it with the latest changes.

Anyway, I added the info banners for errors when adding a member, or to indicate an email was sent in the case of successful pending requests.
Here is what it looks like.
For an error :

For a successful pending request :


Let me know what you both think!

fmigneault:
> > The `Users.status` field could be used for that. Another state could be added to `UserStatuses` (`OK`, `WebhookError`, `Pending` at the moment).
> 
> @fmigneault I am not sure I understand the use of a new UserStatus here. In the case of a fail, when trying to assign a user with a group with T&C, I would just reset the page, making sure not to display a `pending` label for the requested group, removing any related TemporaryTokens, and displaying an error message on the ui.
> 
> The error could come from different causes. For example, an invalid smtp configuration could cause an error, and it is not necessarily something related to the user here, so I don't know if we should keep the error state on the user's status.

The problem with using the page is that unless we verify specifically that this `pending` user-group relation status is defined on the UI page in the memberships list, there would be no way to be aware the user is not fully configured as expected (whether because T&C are not agreed yet or an underlying error). 

Setting the `User.status` with some new value specifically for pending group T&C would report this status in API responses below: 

```json
{
  "user": {
    "user_name": "authtest",
    "email": "authtest@example.com",
    "status": "OK", <--------------- this would indicate something like pending T&C
    "group_names": [
      "anonymous",
      "thredds-secure-authtest-group"
    ],
    "user_id": 4
  },
  "code": 200,
  "detail": "Get user successful.",
  "type": "application/json",
  "path": "/users/authtest",
  "url": "https://host-140-154.rdext.crim.ca/magpie/users/authtest",
  "method": "GET"
}
```

I think it would be valid to have a new status for generic "pending T&C" that are just not yet agreed (eg: case where success status banner you recently added was displayed). This would highlight that the user is not completely "OK" on the API side (as in, it will not behave 100% with the 'final' intention due to pending T&C). 
Maybe a list of `pending_group_names` could also be something to add to this response to help highlight them...


> Anyway, I added the info banners for errors when adding a member, or to indicate an email was sent in the case of successful pending requests.

Just to make sure. The banners should not be displayed if no T&C are to be applied (ie: direct success of user-group link assigned).

magpie/api/management/group/group_views.py
```diff
@@ -116,12 +118,33 @@ def delete_group_view(request):
...
     """
-    List all user from a group.
+    List all users from a group.
+    Users can be filtered by status depending of input arguments.
     """
     group = ar.get_group_matchdict_checked(request)
-    user_names = ax.evaluate_call(lambda: [user.user_name for user in group.users],
-                                  http_error=HTTPForbidden,
-                                  msg_on_fail=s.GroupUsers_GET_ForbiddenResponseSchema.description)
+    status = ar.get_query_param(request, "status", default=UserGroupStatus.ACTIVE.value)
```
fmigneault, line 125: Same as other comment about getting the enum.

magpie/api/management/group/group_views.py
```diff
@@ -116,12 +118,33 @@ def delete_group_view(request):
...
+                    msg_on_fail=s.UserGroup_Check_Status_BadRequestResponseSchema.description,
+                    http_error=HTTPBadRequest)
+
+    user_names = set()
+    member_user_names = ax.evaluate_call(lambda: set(user.user_name for user in group.users),
+                                         http_error=HTTPForbidden,
+                                         msg_on_fail=s.GroupUsers_GET_ForbiddenResponseSchema.description)
+    if status in [UserGroupStatus.ACTIVE.value, UserGroupStatus.ALL.value]:
+        user_names = user_names.union(member_user_names)
+    if status in [UserGroupStatus.PENDING.value, UserGroupStatus.ALL.value]:
```
fmigneault, line 136: Here would also be updated to use enums directly as per other comment.

magpie/api/management/group/group_utils.py
```diff
@@ -75,7 +75,7 @@ def get_group_resources(group, db_session, service_types=None):
     return json_response
 
 
-def create_group(group_name, description, discoverable, db_session):
+def create_group(group_name, description, discoverable, terms, db_session):
     # type: (Str, Str, bool, Session) -> HTTPException
```
fmigneault, line 79: typing missing new param type

magpie/ui/management/views.py
```diff
@@ -626,6 +668,11 @@ def edit_group(self):
         group_info["cur_svc_type"] = cur_svc_type
         group_info["resources"] = res_perms
         group_info["permissions"] = res_perm_names
+
+        if edit_group_users_info:
+            group_info["edit_membership_pending_success"] = \
+                edit_group_users_info["edit_new_membership_success"] & set(group_info["pending_users"])
```
fmigneault, line 674: use variable instead of `\`

tests/test_magpie_ui.py
```diff
@@ -297,6 +297,151 @@ def check_api_resource_permissions(resource_permissions):
...
+                path = "/ui/groups/{}/default".format(group_with_terms_name)
+                resp = utils.test_request(self, "GET", path)
+                body = utils.check_ui_response_basic_info(resp)
+                utils.check_val_not_in("{} [pending]".format(self.test_user_name), body)
+
+                # validate that group membership is no longer pending in the edit user page
+                path = "/ui/users/{}/default".format(self.test_user_name)
+                resp = utils.test_request(self, "GET", path)
+                body = utils.check_ui_response_basic_info(resp)
+                utils.check_val_not_in("{} [pending]".format(group_with_terms_name), body)
```
fmigneault, line 435: Should also check that groups without `[pending]` are now in body?

fmigneault (changes requested):

fmigneault:
@dbyrns @ChaamC 

>     1. Checking a group without T&C should not yield any banner as @fmigneault said **[Done]**
> 
>     2. Checking a group with T&C successfully should show a banner **[Done]** and report this group in a `pending_group_names` section of the user API **[To be done]**
> 
>     3. Checking a group with T&C unsuccessfully should yield an exception banner but not change anything for the user state (the error is reported right away and the user is not added to the group) **[Done]**
> 
>     4. If the point 3 is done via the API, an error response should be returned. Is it the case? **[To validate]**
> 
>     5. The user status should remain unaffected whatsoever. **[Done]**

I agree with all mentioned points. 
I also prefer the (already) added `status` query parameter over `pending_group_names` that was previously mentioned. 

The only thing I believe we are missing is a quick `pending_groups` "status" to show if the user is pending anything when requesting the `GET /users/{user_name}` endpoint. Otherwise, it is not convenient to always have to send 2 requests to fetch details of the user itself, and its pending groups, to know if anything is missing T&C approval. 

I think that this would help managing accounts in the long run if *any* incomplete T&C are reported rapidly in the user details response. Just a summary True/False is enough in this case, details are obtained by `GET /users/{user_name}/groups?pending=True` if needed. The `pending_groups` status could be reported in the user-list UI page such that it is easy to find any user pending T&C, again help manage the multiple accounts.

It is valid to have this field obtained by dynamically checking corresponding TemporaryTokens. It would be great to directly have the method defined in `User` model class. This way, the operation is easily obtainable anywhere needed.

> I think it works, but it would be a good idea to add a test to validate this (a bit like the current `test_PostUserGroupWithTerms` test), which tries to assign a user to a group with T&C, which results in failure. I would then check if no UserGroup relationship was created, and if no temporary tokens were created either.

I agree with that too.

ChaamC:
@fmigneault @dbyrns 
I have added the info about pending_groups to the user info. I haven't added it as a new User status value, since I didn't want to enter in conflict with the WebhookError user status, because a user could have a webhook error while being pending for a group. 

I just added a new boolean field `has_pending_group` in the user info returned by the `GET /users/{user_name}` endpoint.

ChaamC:
@dbyrns @fmigneault 
PR updated according to latest feedback. Let me know if it is okay, or if the latest changes require further adjustments!

Issue #466, opened 2021-09-17
[BUG] Intermittent Twitcher/Magpie error when caching on (Magpie version 3.14.0)


tlvu:
**Describe the bug**
Many random hard to reproduce errors with several services behind Twitcher/Magpie when caching is turned on for Twitcher.

With caching not working, this  issue https://github.com/bird-house/twitcher/issues/97 might have to be re-open?

**To Reproduce**
1. PAVICS at tag https://github.com/bird-house/birdhouse-deploy/tree/1.14.2 (PR https://github.com/bird-house/birdhouse-deploy/pull/188)
2. pavics_thredds.ipynb, stress-test.ipynb in pavics-sdi repo failure
3. Result

===Problem 1===
------------------------
```
11:31:17  _____ pavics-sdi-master/docs/source/notebooks/pavics_thredds.ipynb::Cell 4 _____
11:31:17  Notebook cell execution failed
11:31:17  Cell 4: Cell execution caused an exception
11:31:17  
11:31:17  Input:
11:31:17  # NBVAL_IGNORE_OUTPUT
11:31:17  
11:31:17  AUTH_USR = os.getenv("TEST_MAGPIE_AUTHTEST_USERNAME", "authtest")
11:31:17  AUTH_PWD = os.getenv("TEST_MAGPIE_AUTHTEST_PASSWORD", "authtest1234")
11:31:17  
11:31:17  # Open session
11:31:17  with requests.Session() as session:
11:31:17      session.auth = MagpieAuth(f"https://{PAVICS_HOST}/magpie", AUTH_USR, AUTH_PWD)
11:31:17      # Open a PyDAP data store and pass it to xarray
11:31:17      store = xr.backends.PydapDataStore.open(SECURED_URL, session=session)
11:31:17      ds = xr.open_dataset(store, decode_cf=False)  # Attributes are problematic with this file.
11:31:17  ds
11:31:17  
11:31:17  Traceback:
11:31:17  
11:31:17  ---------------------------------------------------------------------------
11:31:17  HTTPError                                 Traceback (most recent call last)
11:31:17  /tmp/ipykernel_405/1184836852.py in <module>
11:31:17        9     # Open a PyDAP data store and pass it to xarray
11:31:17       10     store = xr.backends.PydapDataStore.open(SECURED_URL, session=session)
11:31:17  ---> 11     ds = xr.open_dataset(store, decode_cf=False)  # Attributes are problematic with this file.
11:31:17       12 ds
[...]

PR #469, opened 2021-09-30
Debug Logging and Attempt to resolve caching issues

fmigneault:
- add more debug logging during effective perm/acl resoluion 
- add session merge/reconnect of `resource` object between service/acl caches

relates to #466

### DIFF
--- a/magpie/adapter/magpieowssecurity.py
+++ b/magpie/adapter/magpieowssecurity.py
@@ -63,10 +63,17 @@ class MagpieOWSSecurity(OWSSecurityInterface):
             - :meth:`magpie.adapter.magpieowssecurity.MagpieOWSSecurity.get_service`
             - :meth:`magpie.adapter.magpieservice.MagpieServiceStore.fetch_by_name`
         """
-        service = evaluate_call(lambda: Service.by_service_name(service_name, db_session=self.request.db),
+        session = self.request.db
+        service = evaluate_call(lambda: Service.by_service_name(service_name, db_session=session),
                                 http_error=HTTPForbidden, msg_on_fail="Service query by name refused by db.")
         verify_param(service, not_none=True, param_name="service_name",
                      http_error=HTTPNotFound, msg_on_fail="Service name not found.")
+        #
+        if cache_regions["service"]["enabled"]:
+            #session.expire(service)
+            #session.refresh(service)
+            session.expunge(service)
+
         # return a specific type of service (eg: ServiceWPS with all the ACL loaded according to the service impl.)
         service_impl = service_factory(service, self.request)
         service_data = dict(service.get_appstruct())
@@ -87,6 +94,7 @@ class MagpieOWSSecurity(OWSSecurityInterface):
         if "service" not in cache_regions:
             cache_regions["service"] = {"enabled": False}
         if self.request.headers.get("Cache-Control") == "no-cache":
+            LOGGER.debug("Cache invalidation requested. Removing items from service region: [%s]", service_name)
             invalidate_service(service_name)
 
         # retrieve the implementation and the service data contained in the database entry
--- a/magpie/adapter/magpieowssecurity.py
+++ b/magpie/adapter/magpieowssecurity.py
@@ -90,6 +90,7 @@ class MagpieOWSSecurity(OWSSecurityInterface):
             invalidate_service(service_name)
 
         # retrieve the implementation and the service data contained in the database entry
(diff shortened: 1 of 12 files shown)


fmigneault:
@tlvu 
Once validated, I will tag a new patch version with added logs. 
You can try to see if the added session merge operation resolves your issue. 

When re-testing, see if you can enable DEBUG logging for both Magpie and Twitcher so we can have a clear idea of which steps are happening in case the caching-related error still occurs.

If it does happen, I would also appreciate if you could try running Twitcher with `gunicorn` instead of `waitress`. 
I can see your logs in #466 are formed as follows (`waitress`): 
```log
2021-09-09 15:24:02,563 ERROR [TWITCHER:33][waitress-0] ...
```
While mine are as follows (`gunicorn`): 
```log
2021-09-29 22:43:58,303 DEBUG [TWITCHER:177][ThreadPoolExecutor-0_1]  ...
```

I'm not sure if their worker/threading vs memory sharing strategies are equivalent, but I can never replicate the error you get when using `gunicorn`. Because `waitress` only reports `-0` rather than `0_1` as `gunicorn` does, it have a feeling their multiprocessing are handled differently, which could impact how in-memory caches work between parallel requests.

tlvu:
> I would also appreciate if you could try running Twitcher with `gunicorn` instead of `waitress`.

Isn't this choice in the Twitcher docker image, the CMD or ENTRYPOINT line?

If in birdhouse-deploy, please make a PR and I can enable it on our staging server and try another round of Jenkins run to smoke out the problem.

Have you been able to reproduce it from a real PAVICS stack on your end?  Otherwise, you can hit our staging server https://medus.ouranos.ca.  I got it reproduced a few times.

magpie/adapter/magpieowssecurity.py
```diff
@@ -155,7 +162,8 @@ def check_request(self, request):
                                  service_impl.service.resource_name, type(perm_exc).__name__, perm_exc)
 
             if permission_requested:
-                LOGGER.info("'%s' request '%s' permission on '%s'", request.user, permission_requested, request.path)
+                LOGGER.info("User %s is requesting '%s' permission on [%s]",
```
tlvu, line 165: Is Twitcher and Magpie container python 3.6 and newer?  If yes, you can use f-string that is way more readable.

That log line with f-string will look like `LOGGER.info(f"User {request.user} is requesting '{permission_requested}' permission on [{request.path}]"`

tlvu (approved):
LGTM, ignore my f-string comment if compat with python 3.5 and below is needed or if any other reasons.  This is not blocking at all.

fmigneault:
> Isn't this choice in the Twitcher docker image, the CMD or ENTRYPOINT line?

The docker offers both `waitress` and `gunicorn`. It is selected based on the INI file. 
The default Twitcher configuration uses `gunicorn` by default since `v0.5.4` (Magpie now points at `v0.5.6`). 
https://github.com/bird-house/twitcher/blob/0.5.x/CHANGES.rst#054-2020-10-29
[`v0.5.3...0.5.x`#diff-28b1353404](https://github.com/bird-house/twitcher/compare/v0.5.3...0.5.x#diff-28b1353404cfe6407a1e205f7e976b53023ceb27f0e86e499090b937316b8c84R46-R49)

It is overridden in birdhouse-deploy: 
https://github.com/bird-house/birdhouse-deploy/blob/master/birdhouse/config/twitcher/twitcher.ini.template#L83-L85

I have not been able to test on a full stack. I will push the tag now to make the patch available.
I will be unavailable most of tomorrow so you can give it a try first and not have wait after me.

tlvu:
> I have not been able to test on a full stack. I will push the tag now to make the patch available.

Sorry, I was not clear.  I meant reproduce the faulty behavoir with the current old version of Magpie/Twitcher that is in birdhouse-deploy right now but on the full PAVICS on your side.

> I will be unavailable most of tomorrow so you can give it a try first and not have wait after me.

No rush.  I am working on Geoserver this week.  We had a power outage at the beginning of the week with problems so I already lost some time there.

magpie/adapter/magpieowssecurity.py
fmigneault: It's always better to have `%s` with extra args in logging calls because it skips the formatting step entirely when log level is not reached. This adds-up quickly in term of delays if there are a lot of log calls.

tlvu: Oh nice, I have not thought about this performance optimization side.

PR #471, opened 2021-10-01
add type query parameter to filter requested services/resource tree

### LINKED ISSUE #463: [BUG] Bad effective user permission display in UI
**Describe the bug**
Bad effective user permission display in UI only, when executing notebook {pavics-sdi repo}/docs/source/notebooks/pavics_thredds.ipynb it works fine.

**To Reproduce**
1. Deploy PAVICS stack at tag `1.14.0` (Magpie `3.14.0`) with `optional-components/secure-thredds` enabled
2. Execute `scripts/create-magpie-authtest-user`
3. Go to https://PAVICS_HOST/magpie/ui/users/authtest/thredds
4. Check "View inherited group permissions and effective user permissions. "

**Expected behavior**
`authtest` user effective permission should have been Allow, not Denied as in screenshot.

**Screenshots**
Screenshot 2021-08-26 at 14-26-51 Magpie

fmigneault:
# Changes

* Add ``type`` query parameter to multiple requests returning ``Services`` or ``Resources`` regrouped by ``ServiceType``
  for a given ``User`` or ``Group`` in order to limit listing in responses and optimise some operations where only a
  subset of details are needed.
* Using new ``type`` query to filter ``ServiceType``, improve ``Permissions`` listing in UI pages with faster processing
  because ``Services`` that are not required (since they are not currently being displayed by the tab-panel view) can
  be skipped entirely, removing the need to compute their underlying ``Resource`` and ``Permissions`` tree hierarchy.

# References

- fixes #463

### DIFF
--- a/magpie/api/schemas.py
+++ b/magpie/api/schemas.py
@@ -505,8 +505,8 @@ QueryFlattenServices = colander.SchemaNode(
 QueryFilterServiceType = colander.SchemaNode(
     colander.String(), name="type", missing=colander.drop,
     example="api,thredds",
-    description="Comma-separated list of service-type for which to obtain descriptions instead of all available ones. "
-                "Provided types are case insensitive. Unknown types are ignored.")
+    description="Comma-separated list of service-type for which to filter retrieved descriptions instead of all "
+                "available ones. Provided types matching is case insensitive. Unknown types are ignored.")
 
 
 class PhoenixServicePushOption(colander.SchemaNode):
@@ -2277,6 +2277,7 @@ class UserServices_GET_QuerySchema(QueryRequestSchemaAPI):
     cascade = QueryCascadeResourcesPermissions
     inherit = QueryInheritGroupsPermissions
     flatten = QueryFlattenServices
+    svc_type = QueryFilterServiceType
 
 
 class UserServices_GET_RequestSchema(BaseRequestSchemaAPI):
@@ -2499,8 +2500,13 @@ class GroupUsers_GET_ForbiddenResponseSchema(BaseResponseSchemaAPI):
     body = ErrorResponseBodySchema(code=HTTPForbidden.code, description=description)
 
 
+class GroupServices_GET_QuerySchema(QueryRequestSchemaAPI):
+    svc_type = QueryFilterServiceType
+
+
 class GroupServices_GET_RequestSchema(BaseRequestSchemaAPI):
     path = Group_RequestPathSchema()
+    querystring = GroupServices_GET_QuerySchema()
 
 
 class GroupServices_GET_ResponseBodySchema(BaseResponseBodySchema):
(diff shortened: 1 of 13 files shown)


magpie/api/schemas.py
```diff
@@ -479,13 +479,15 @@ class QueryRequestSchemaAPI(colander.MappingSchema):
...
     description="Include the user's groups memberships inheritance to retrieve all possible permissions. "
-                "(Note: Duplicate, redundant or conflicting permissions can be obtained considering applied group"
-                " permissions individually. See 'combined' query parameter to resolve such cases.)")
-QueryCombinedGroupsPermissions = colander.SchemaNode(
-    colander.Boolean(), name="combined", default=False, missing=colander.drop,
-    description="Combines corresponding user and groups inherited permissions into one, and locally resolves "
-                "for every resource the applicable permission modifiers considering group precedence. "
-                "(Note: Group inheritance is enforced regardless of 'inherited' query parameter.)")
+                "(Note: Duplicate, redundant and even conflicting permissions can be obtained when considering "
+                "multiple applied permissions individually applied fro different group memberships or for the user "
```
dbyrns, line 483: typo fro -> from

magpie/ui/management/views.py
```diff
@@ -465,11 +465,17 @@ def get_user_or_group_resources_permissions_dict(self, user_or_group_name, servi
               or :term:`Group` accordingly to specified arguments.
         """
         if is_user:
-            query = "?inherited=true" if is_inherit_groups_permissions else ""
-            path = schemas.UserResourcesAPI.path.format(user_name=user_or_group_name) + query
+            # because page can only show a single permission (per name/resource) at a time, apply resolution
```
dbyrns, line 467: What's the difference with before?
The single permission shown should be the effective, no?

dbyrns (approved):
I'm wondering what is changing exactly in the shown permission and found a typo. Else everything looks good.

magpie/ui/management/views.py
fmigneault: No. The displayed permission is either direct or inherited in the combobox, based on the checkbox "show intherited". 
The effective permission is obtained by the test button next to it that transforms into checkmark/cross. 

The problem was that given a resource X, a user member of both groups A and B, each with different permissions on X was not resolving the most important one to be displayed. 
Cases where `[(user, X, allow-match), (A, X, allowed-recurive), (B, X, deny-match)]` are applied would not be resolved and would simply show the last item in the HTML `<select> <options>` (that's how it resolves which one to show when many are selected for single-select comboboxes).

dbyrns: Hum that's right, we don't have a "word" for the "most important one". But I meant the "local effective", without considering the scope. Anyway, we now have the good behavior. Thank you!

fmigneault: Indeed there is no 'term' in the glossary, but I refer to that concept as `resolved` in docs.
It is present here: https://pavics-magpie.readthedocs.io/en/latest/permissions.html#types-of-permissions

PR #484, opened 2021-11-25
Fix reconnect user / session

fmigneault:
Using test results of https://github.com/bird-house/birdhouse-deploy/pull/197#issuecomment-978164867
And the extended stress-test logs in daccs-jenkins.crim.ca/job/PAVICS-e2e-workflow-tests/job/master/663

We can see that previous places where most of the time that ``DetachedInstanceError`` and ``NoTransaction`` were raised are gone. 
Only two case where raised within the thousands of requests.

Attempt to fix them with similar operation reconnecting the session or reattaching the object to it. 
Added a minimal test to early detect such situations in the future if a new step is added to the code, but those errors are really hard to evaluate more than this.


dbyrns (approved):
LGTM

PR #488, opened 2021-11-27
Retry session tm failure

fmigneault:
## Overview

Even with previous patches in `3.18.x` releases, some cases still seem to generate sporadic `NoTransaction` errors. 

## Changes 

- In order to (hopefully) fix this once and for all, add a retry operation specifically for the corresponding errors happening when desync of cached vs live db session are used.
- Combine the transaction/database/user configuration attached to the request on both `Magpie` itself and the `MagpieAdapter`, into a reusable setup function. 
- Clean/refactor the config setup using above function and centralize them into application main.
- Attempt session reconnect in a few more places that where highlighted in https://github.com/bird-house/birdhouse-deploy/pull/197#issuecomment-980431919
- Integrate fixes from https://github.com/bird-house/twitcher/pull/111

### DIFF
--- a/.github/.gitleaks.toml
+++ b/.github/.gitleaks.toml
@@ -115,8 +115,4 @@ title = "gitleaks config"
     '''(go.mod|go.sum)$''',
     # old commit files with false positives or dummy data
     '''magpie/login/login.py''',
-<<<<<<< HEAD
     '''.+(.js.map)$''']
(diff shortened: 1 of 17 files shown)


dbyrns (dismissed):
LGTM

dbyrns (dismissed):

PR #490, opened 2021-12-03
add log entry to investigate failing service creation

fmigneault:
## Description

In some very rare/sporadic cases, the service that gets created from scratch using a request that just followed cache reset in the following part causes an unhanded error (see referenced issue).
https://github.com/Ouranosinc/Magpie/blob/87fe4d70a8836d609711ff8819e4aaeb79f3b02c/magpie/adapter/magpieowssecurity.py#L85

Add some logs in order to investigate the cause further. 
Since this really doesn't happen often (only happened once at the time of writing following at least 3 stress-test successful runs), the only thing we can do it try to detect it again and fix accordingly when we obtain more details.

## References

- relates to https://github.com/bird-house/birdhouse-deploy/pull/224#issuecomment-985668339

### DIFF
--- a/magpie/services.py
+++ b/magpie/services.py
@@ -973,14 +973,23 @@ def service_factory(service, request):
     """
     ax.verify_param(service, param_compare=models.Service, is_type=True,
                     http_error=HTTPBadRequest, content={"service": repr(service)},
-                    msg_on_fail="Cannot process invalid service object")
+                    msg_on_fail="Cannot process invalid service object.")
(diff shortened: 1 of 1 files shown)


fmigneault:
Seen again during https://github.com/bird-house/birdhouse-deploy/pull/225#issuecomment-985740171

dbyrns (commented):
I performed a quick review anyway and everything looks good.

PR #495, opened 2022-01-10
Implement combined Geoserver service and Workspace/Layer effective permissions

fmigneault:
## Summary

First implementation that resolves most common use cases. 
Since the changes are already substantial, start with this version that **DOES NOT** support multiple `layers=<layer1>,<layer2>` resources. 

A subsequent PR can attempt resolution of combined resources and return only `Allow` if all resources are accessible. 
That will involve modifying the (complicated) `effective_permissions` method.

## Features / Changes

* Add missing ``ServiceWFS`` permissions according to [OGC WFS standard](https://www.ogc.org/standards/wfs).
* Add missing ``DescribeLayer`` permission to ``ServiceGeoserverWMS`` according to [GeoServer WMS implementation](https://docs.geoserver.org/latest/en/user/services/wms/reference.html).
* Add support of specific hierarchy of ``Resource`` type ``Layer`` nested under ``Workspace`` for ``ServiceGeoserverWMS``.
* Add support of ``Resource`` type ``Layer`` under ``ServiceWFS``.
* Allow ``Resource`` and ``Service`` name to contain colon (``:``) character in order to define scoped names as it is often the case for ``Layer`` names.
* Add ``child_structure_allowed`` attribute to ``Service`` implementations allowing them to define specific path-like structures of allowed ``Resource`` types hierarchies in order to control at which level and which combinations of nested ``Resource`` types are valid under their root ``Service``. When not defined under a ``Service`` implementation, any defined ``Resource`` type will remain available for creation at any level of the hierarchy, unless the corresponding ``Resource`` in the tree already defined ``child_resource_allowed = False``. This was already the original behaviour in previous versions.
* Add ``GET /resources/{id}/types`` endpoint that allows retrieval of applicable children ``Resource`` types under a given ``Resource`` considering the nested hierarchy definition of its root ``Service`` defined by the new attribute ``child_structure_allowed``.
* Add ``child_structure_allowed`` attribute to the response of ``GET /service/{name}`` endpoint.
  For backward compatibility, ``resource_types_allowed`` parameter already available in the same response will continue to report all possible ``Resource`` types *at any level* under the ``Service`` hierarchy, although not necessarily applicable as immediate child ``Resource`` under that ``Service``.
* Add ``configurable`` attribute to ``Service`` types that supports custom definitions modifying their behaviour.
* Add ``service_configurable`` to response of ``GET /service/{name}`` endpoint.
* Adjust UI to consider ``child_structure_allowed`` definitions to propose only applicable ``Resource`` types in the combobox when creating a new ``Resource`` in the tree hierarchy.
* Add UI submission field to provide ``Service`` JSON configuration at creation when supported by the type.

## Bug Fixes

* Remove invalid ``params_expected`` parameter from ``Service`` implementations (``ServiceAccess``, ``ServiceAPI``,
  ``ServiceTHREDDS``) that don't make use of it since they don't derive from ``ServiceOWS``.
* Fix base ``Permission`` definitions for all variants of `WMS` according to their reference implementations.
* Remove multiple invalid schema path definitions that are not mapped against any concrete API endpoint.
* Fix reporting of ``Service`` configuration for any type that supports it. Unless overridden during creation with a custom configuration, ``ServiceTHREDDS`` implementation would not report their default configuration and would instead return ``null``, making it difficult to know from the API if default or no configuration was being applied for a given ``Service``.
* Fix `Effective Resolution` of ``Permission`` applied for ``ServiceGeoserverWMS`` to consider ``Scope`` modifier of ``Service`` and ``Workspace`` for access to be resolved at the ``Layer`` level.

### DIFF
--- a/docs/services.rst
+++ b/docs/services.rst
@@ -174,7 +174,7 @@ attempting access of a specific :term:`Resource` of type :class:`magpie.models.F
     Permission :attr:`Permission.READ` does not offer *metadata* content listing of :class:`magpie.models.Directory`
     anymore. For this, :attr:`Permission.BROWSE` should be used instead. Setting :attr:`Permission.READ` on a
     directory will only be logical when combined with :attr:`Scope.RECURSIVE`, in which case `Magpie` will interpret
-    the :term:`Effective Permissions` to allow read access to all :class:`magpie.models.File` under that directory, at
+    the :term:`Effective Permission` to allow read access to all :class:`magpie.models.File` under that directory, at
     any depth level, unless denied by a lower-level specification.
 
 Finally, :attr:`Permission.WRITE` can also be applied on all of the resources, but are not explicitly employed during
@@ -186,7 +186,7 @@ parsing of incoming requests.
     only returns either :attr:`Permission.BROWSE` or :attr:`Permission.READ`. A :term:`User` or :term:`Group` can still
     have this :term:`Applied Permission` to allow a third party service to interrogate `Magpie API` about the presence
     of :attr:`Permission.WRITE` permission and perform the appropriate action with the result. The
-    :term:`Effective Permissions` API routes will provide the resolved ``access``. It is only `Twitcher`_ proxy that
+    :term:`Effective Permission` API routes will provide the resolved ``access``. It is only `Twitcher`_ proxy that
     will not be able to make use of it during incoming requests as it depends on
     :class:`magpie.adapter.magpieowssecurity.MagpieOWSSecurity`, which in turn employs the result from the :term:`ACL`.
 
@@ -312,7 +312,7 @@ To summarize, if ``file_patterns`` produces a match, that matched portion will b
 name is used directly (as is from the specified request path). The plain name is also used if ``file_patterns`` is
 explicitly specified as an empty list or ``null``. Not explicitly overriding the field will result into using the
 above *default* ``file_patterns``. The ``file_patterns`` allow for example to consider ``file.nc``, ``file.ncml`` and
-``file.nc.html`` as the same :term:`Resource` internally, which avoids duplicating :term:`Applied Permissions` across
+``file.nc.html`` as the same :term:`Resource` internally, which avoids duplicating :term:`Applied Permission` across
 multiple :term:`Resource` for their corresponding *metadata* or *data* representations.
 
 
--- a/docs/services.rst
+++ b/docs/services.rst
@@ -78,8 +78,8 @@ On top of the above methods, the following attributes must be defined.
         :Term:`Allowed Permissions`, their type and further nested children :term:`Resource`.
     * - :attr:`ServiceInterface.params_expected` |br| (``List[str]``)
       - Represents specific parameter names that can be preprocessed during HTTP request parsing to ease following
-        resolution of :term:`ACL` use cases. Employed most notably by :term:`Service` implementations based on
-        :class:`magpie.services.ServiceOWS`.
+        resolution of :term:`ACL` use cases. Employed only by :term:`Service` implementations derived from
+        :class:`magpie.services.ServiceOWS`. Can be omitted otherwise.
 
 
 .. _services_available:
--- a/docs/services.rst
+++ b/docs/services.rst
@@ -32,14 +32,15 @@ Every :term:`Service` type provided by `Magpie` must derive from :class:`magpie.
 specific implementation (see :ref:`services_available`) serves to convert a given incoming HTTP request components
 (method, path, query parameters, body, etc.) into the appropriate :term:`Service`, :term:`Resource` and
 :term:`Permission` elements. This ultimately provides the required elements to resolve :term:`ACL` access of a
-:term:`Request User` toward the targeted :term:`Resource` according to its :term:`Effective Permissions`.
+:term:`Request User` toward the targeted :term:`Resource` according to its
+:term:`Effective Permissions <Effective Permission>`.
 
 In order to implement a new :term:`Service` type, two (2) methods and a few attributes are required. The first method is
 :meth:`magpie.services.ServiceInterface.permission_requested` which basically indicates how the HTTP request should be
 interpreted into a given :class:`Permission`. The second is :meth:`magpie.services.ServiceInterface.resource_requested`
 which similarly tells the interpretation method to convert the request into a :class:`magpie.models.Resource` reference.
 
-Whenever :term:`Effective Permissions` or :term:`ACL` needs to be resolved in order to determine if a
+Whenever :term:`Effective Permission` or :term:`ACL` needs to be resolved in order to determine if a
 :term:`Request User` can have access or not to a :term:`Resource`, `Magpie` will employ the appropriate :term:`Service`
 implementation and call the methods to process the result.
 
@@ -79,7 +80,7 @@ On top of the above methods, the following attributes must be defined.
         also provides details about :Term:`Allowed Permissions <Allowed Permission>`, their type and further nested
         children :term:`Resource`.
     * - :attr:`child_structure_allowed` |br| (``Dict[Type[ServiceOrResourceType], List[Type[models.Resource]]]``)
-      - Map of allowed :term:`Resource`type nesting hierarchy for the :term:`Service`.
+      - Map of allowed :term:`Resource` type nesting hierarchy for the :term:`Service`.
         This controls whether some children :term:`Resource` can be placed under another to limit creation only to
         cases that are relevant for the implemented :term:`Service`.
     * - :attr:`configurable` |br| (``bool``)
@@ -176,9 +177,9 @@ ServiceAPI
 The implementation of this :term:`Service` is handled by class :class:`magpie.services.ServiceAPI`. It refers to a
 remote URL endpoint that should have a :term:`Resource` tree formed out of the path segments. The :term:`Service` only
 has one (1) type of :term:`Resource`, namely :class:`magpie.models.Route`, that can have an unlimited amount of nested
-children of the same type. The :term:`Allowed Permissions` for this :term:`Service` are :attr:`Permission.READ` and
-:attr:`Permission.WRITE`. All requests using ``GET`` or ``HEAD`` are mapped to :attr:`Permission.READ` access while all
-other HTTP methods represent a :attr:`Permission.WRITE` access.
+children of the same type. The :term:`Allowed Permissions <Allowed Permission>` for this :term:`Service`
+are :attr:`Permission.READ` and :attr:`Permission.WRITE`. All requests using ``GET`` or ``HEAD`` are mapped
+to :attr:`Permission.READ` access while all other HTTP methods represent a :attr:`Permission.WRITE` access.
 
 Request path segments follow the natural hierarchy of the nested :class:`magpie.models.Route` under the :term:`Service`.
 For example, a proxy employing :class:`magpie.adapter.MagpieAdapter` such that ``{PROXY_URL}`` is the base of the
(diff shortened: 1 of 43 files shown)


dbyrns (dismissed):
I got some quick fixes, but overall this is a good PR. 
@f-PLT, since I only did a static review, you should validate the implementation with a live geoserver instance so that it fits correctly our use cases before approving.
Thank you!

magpie/services.py
```diff
@@ -60,19 +68,82 @@ def resource_type_names(cls):
...
+    created at some point before the last ``process`` can even exist, partial paths (without ``process``) must also
+    be allowed as valid structures.
+
+    .. code-block:: python
+
+        child_structure_allowed = [
+            "service/route",
+            "service/route/process",
+            "service/route/route",
+            "service/route/route/process",
```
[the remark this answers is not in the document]
fmigneault: The regex approach is exactly what I tried at first. ;) 
It looked like `service(/route)+(/process)?` to allow the route creation before `process` can exist. 

I encountered a problem though when trying to implement `resource_requested` because I somehow need to guess the "next type" from the path to provide which one are applicable during child resource creation, which became very complex with "any regex" form.

magpie/services.py
```diff
@@ -322,11 +393,80 @@ def get_resource_permissions(cls, resource_type_name):
...
+        else:
+            res_impl = models.RESOURCE_TYPE_DICT[parent_resource.resource_type_name]
+        if not res_impl.child_resource_allowed:
+            return False
+        # if undefined control structures, assume any combination of nested resource is allowed (original behaviour)
+        if not cls.child_structure_allowed:
+            return True
+        res_types_path = cls.get_resource_type_path(parent_resource, extra_path=child_resource_type)
+        for allow_types_path in cls.child_structure_allowed:
+            if allow_types_path == res_types_path:
```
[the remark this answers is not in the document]
fmigneault: Yes. That is the intended behaviour. It enforces very explicit structures only. To allow `"service/toto/tata"`, it must be added to the list (as well as all its intermediate paths to allow their creation). It is an "extra" control over which resources can be created or not.

When `child_structure_allowed` is unspecified (as in `ServiceAPI` for example), you can create any resource in `resource_types_permissions` (`Route` for that service) at any nested level as long as they also have `child_resource_allowed = True`.

magpie/services.py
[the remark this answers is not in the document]
dbyrns: Could it be simply resolved by defining the allowed child by resource type?
```
allowed_child_resource_type = {
   "service" : ["directory"],
   "directory" : ["directory", "file"],
   "file": []
}
allowed_child_resource_type = {
   "service" : ["route"],
   "route" : ["route", "process"],
   "process" : []
}
allowed_child_resource_type = {
   "service" : ["workspace"],
   "workspace" : ["layer"],
   "layer" : [] 
}
```
Or at some point you need to control the child type based on the current depth?

magpie/services.py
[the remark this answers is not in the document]
dbyrns: I'm probably looking for too much flexibility for the use case 😉. Let's conclude on the other thread regarding child_structure_allowed.

magpie/services.py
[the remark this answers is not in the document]
fmigneault: It could work, except if we encounter a case were very specific nested placement is required. 
I'm thinking for example of  `service/route/route/route/leaf`, where there must be exactly 3 `route` items and only the last can have the `leaf` item.
Technically we could work around that by defining different classes for each of those items.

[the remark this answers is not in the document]
dbyrns: Then, from my perspective, I find it neater to define a child relation, but if you want to merge this PR as is, I will not push for it.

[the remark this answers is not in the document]
fmigneault: Given that I am currently creating the "paths" using format in the following manner, I tend to agree that using dicts like you propose would be more readable. 

```python 
"{}/{}/{}".format(
            models.Service.resource_type_name,
            models.Workspace.resource_type_name,
            models.Process.resource_type_name,
        )
```

I'll give it a try.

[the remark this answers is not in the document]
fmigneault: Applied in 9c298f5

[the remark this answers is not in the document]
dbyrns: Nice job! It looks like the config and the code handling it is cleaner.

f-PLT:
Tested with the public urls on a test instance and everything works very well and as expected with workspaces and layers.

I did find a small problem with WFS request parameters: 
`typeNames (version=2.0.0) vs typeName (version<=1.1.0)`

We'll have to account for this either by allowing/filtering for both, whatever the WFS version request, or filter by WFS version.

However, Geoserver doesn't seem to care and will accept both for a request using version 2.0.0, which leads to the following security bypass:

Using the following configuration should give access to the workspace `isolated`, but restrict access to the `routes` layer

image

- _HOST_URL/wfs?&version=2.0.0&request=GetFeature&**typeNames**=isolated:routes_
will return `Access to service is forbidden`, as expected 

- _HOST_URL/wfs?&version=2.0.0&request=GetFeature&**typeName**=isolated:routes_ 
will be successful

fmigneault:
Self-notes of things to adjust following feedback and discussions :
- [ ] modify `resource_param` to support both `typename(s)` variants indiscriminately 
- [ ] add test validating use case where deny permission revokes access on Layer from allowed Workspace 
- [ ] adjust `GET /services/types/{}/resources` to return contextual `resource_types_permissions` instead of the full service's permissions
- [ ] adjust services `resource_types_permissions` to apply relevant permissions (eg: GetCapabilities valid on WFS service, but not on Workspace and Layer)
- [ ] adjust UI to consider above and disable/grey-out non applicable permissions on resources by type
- [ ] realign permission title with comboboxes (ok on Firefox, but not on Chrome)
- [ ] consider http://jsfiddle.net/kunknown/VVaEq/2/ to horizontally scroll permissions titles/bboxes divs together /w max with

docs/services.rst
```diff
@@ -424,6 +515,80 @@ it against an existing :term:`Resource`. The resolution of :term:`Effective Perm
...
+    providers:
+      RemoteGeoServer:
+        # minimal configuration requirements (where the real `GeoSever` service resides)
+        # other optional parameters from `providers.cfg` can also be provided
+        url: http://localhost:1234
+        type: geoserver
+
+        # customizable configuration (enable desired OWS request handlers)
+        # all OWS are enabled by default if no configuration is provided
+        configuration:
```
dbyrns, line 572: Is it new that service type can have custom config? Does the UI allow to edit that?

magpie/services.py
```diff
@@ -287,28 +309,63 @@ def _get_acl_cached(self, service_name, request_method, request_path, user_id):
...
         """
-        Resolves the resource-tree and the user/group inherited permissions into a simplified ACL for this resource.
+        Resolves resource-tree and user/group inherited permissions into simplified :term:`ACL` of requested resources.
+
+        Contrary to :meth:`effective_permissions` that can be resolved only for individual :term:`Resource`, :term:`ACL`
+        is involved during actual request access, which could refer to multiple :term:`Resource` references or distinct
+        :term:`Permission` names if the identified parent :term:`Service` supports it.
+
+        When more than one item is specified for validation (any combination of :term:`Resource` or :term:`Permission`),
+        **ALL** of them must be granted access to resolve as :data:`Access.ALLOW`. Any denied access blocks the whole
```
dbyrns, line 342: So you're allowing to deny a wfs or a wms and this will result in denying access to both?
This is kind of adding a whole new dimension to this inheritance|allow|deny|group membership puzzle 😜. How this decision is reflected in the UI? In the API reason?

magpie/services.py
```diff
@@ -844,57 +913,119 @@ class ServiceGeoserverBase(ServiceOWS):
...
+            instead of nesting ``<namespace>`` and ``<element>`` into two distinct :term:`Resources <Resource>`.
         """
         raise NotImplementedError
 
-    @property
+    @classproperty
     @abc.abstractmethod
-    def resource_param(self):
-        # type: () -> Str
+    def resource_multi(cls):  # noqa  # pylint: disable=E0213,no-self-argument
```
dbyrns, line 940: Maybe my guess about wfs/wms was bad. Are we rather checking a request that involve multiple resources at the same time and thus all of them must be allowed or the request is deny. In which case, and it makes sense, the permission UI is not really affected.

dbyrns (commented):
Got some questions along the road, but it looks really good!

docs/services.rst
fmigneault: No(1) and yes(2). 
Before, only `ServiceTHREDDS` had a customizable config (what defines meta/data access for `browse`/`read` permissions).
Both POST/PATCH requests allow a JSON in `configuration` field. They are then padded with defaults in case of missing contents.

magpie/services.py
fmigneault: No, because only one between WFS/WMS can be targeted. The service first resolves `geoserver/<OWS>`, and then the permisions applicable for the identified `<OWS>`. Also, they would most probably not have the same permission derived from `request=GetMap|GetFeature|...` parameter.

This **ALL(Allow)** concept is more about `layers=layer1,layer2`. Both must be allowed to allow the full request.

magpie/services.py
fmigneault: Yes, that's exactly it.
Technically the code allows many resource combinations across different services, but each service is usually selected and returns its own children resources.

docs/services.rst
fmigneault: Will need to add the option to edit via UI. 
It is possible to define it during service creation, but not modify it afterward.

image

dbyrns: Ok, it doesn't have to be in that PR though.

magpie/services.py
dbyrns: Sorry, I should have resolved this thread once I added the following comment.

dbyrns (approved):
Great job! Do you think that at some point `magpie/services.py` could be split? It starts to contains a lot of stuff.

docs/services.rst
fmigneault: Yes, I thought so too : #500

fmigneault:
@dbyrns 
> Do you think that at some point `magpie/services.py` could be split?

Could be but I don't see any advantage other than to reduce how long the file is.
There would need to be many indirect imports to ensure references still work (eg: `SERVICE_TYPE_DICT` imported from many spaces, and itself importing from many places...)

f-PLT:
I redid my tests for `GetFeature` and everything works well on my end.

I tested `GetCapabilities` a bit more in depth, directly on the workspace url (ex. _host.domain/geoserver-service-public-url/workspace-name_)

Whether I set the permission to Allow or Deny at the workspace level, the permission at the service level seems to take precedence.

image
With the above permissions, I still get an _Access is forbidden_.

With **Service** => _Allow_ and **Workspace** => _Deny_, I can still make a `GetCapabilities` request on the specific workspace.

fmigneault:
@f-PLT 
Would be worth testing another permission because `GetCapabilities` is handled differently. 
It makes senses only onto the service itself.
The service implementation stops even considering any other resource than the service itself in that case.

dbyrns:
Looking at the screenshot posted by @f-PLT, it looks like the permissions don't line up properly with the comboboxes. Is it because of total number of permissions or because some of them seem to explose the allowed width (horizontal comboboxes)?

f-PLT:
> Looking at the screenshot posted by @f-PLT, it looks like the permissions don't line up properly with the comboboxes. Is it because of total number of permissions or because some of them seem to explose the allowed width (horizontal comboboxes)?

Already has been made an issue by @fmigneault  : https://github.com/Ouranosinc/Magpie/issues/498

fmigneault:
@dbyrns 
Seems to be a problem in Chrome. I actually noticed only with screenshots by @f-PLT because they line up correctly in Firefox.

image

When there are many permissions though, the UI needs to be improved also because there no clean way to display them, lined-up or not (they wrap around on 2nd line).

image

Issue #498, opened 2022-01-28
Update UI to handle many permissions wider than available display width


fmigneault:
- [x] realign permission title with comboboxes (ok on Firefox, but not on Chrome)
- [x] consider http://jsfiddle.net/kunknown/VVaEq/2/ to horizontally scroll permissions titles/bboxes divs together /w max width

_Originally posted by @fmigneault in https://github.com/Ouranosinc/Magpie/pull/495#issuecomment-1019011015_

fmigneault:
resolved by #502

PR #501, opened 2022-02-01
fix alignment of permission title/combobox

fmigneault:
## Changes / Fixes

* Add auto-restore of previous scroll position in UI page following submitted form.
* Add UI tooltip `Resource` ID to elements rendered in the ``Service`` and ``Permission`` hierarchy trees
  (relates to #335).
* Fix UI alignment of permission titles with their corresponding permission selectors
  (relates to #498).

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -56,7 +56,8 @@ Bug Fixes
   for a given ``Service``.
 * Fix `Effective Resolution` of ``Permission`` applied for ``ServiceGeoserverWMS`` to consider ``Scope`` modifier
   of ``Service`` and ``Workspace`` for access to be resolved at the ``Layer`` level.
-* Fix UI alignment of permission titles with their corresponding permission selectors.
+* Fix UI alignment of permission titles with their corresponding permission selectors
+  (relates to `#498 <https://github.com/Ouranosinc/Magpie/issues/498>`_).
 
 `3.20.1 <https://github.com/Ouranosinc/Magpie/tree/3.20.1>`_ (2022-01-19)
 ------------------------------------------------------------------------------------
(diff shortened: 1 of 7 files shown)


dbyrns (commented):
It's a nice addition to store the scrolling position.
If my understanding is correct the tooltip apply to the ressources, but it would be really usefull for the permission title too when there is an overflow and ellipsis is applied.
I will let @f-PLT makes the final approval after he checks the rendering.

f-PLT:
Code looks good to me, and rendering worked as expected on Firefox and Chrome on my end.

f-PLT (approved):

fmigneault:
@dbyrns 
> for the permission title too when there is an overflow and ellipsis is applied.

Doesn't it display a scrollbar when the title is too large? That's what it is supposed to do.

dbyrns:
> Doesn't it display a scrollbar when the title is too large? That's what it is supposed to do.

Didn't test it. `text-overflow: ellipsis;` for me is used when `...` replace the text that don't fit it the space. Fine with me if a scrollbar is available and all the permission can be read.

dbyrns (approved):

f-PLT:
> > Doesn't it display a scrollbar when the title is too large? That's what it is supposed to do.
> 
> Didn't test it. `text-overflow: ellipsis;` for me is used when `...` replace the text that don't fit it the space. Fine with me if a scrollbar is available and all the permission can be read.

I did a quick test with CSS in an editor, the lack of `...` could be related to `  display: inline-flex;` in the `permission-cell` class, as they are applied to the same divs. 

Actually replacing with `...` could make it confusing for people who are not familiar with the operations in question (unless a tooltip or other way to see the full name is available)

Anyway, I think that once a full length horizontal scrollbar is available for permission (as is cited in https://github.com/Ouranosinc/Magpie/issues/498), the columns could also be widened so the full title is visible for all columns.

fmigneault:
merging as is, lint errors resolved in PR #502

fmigneault:
For documentation purpose, the tooltip in question: 

image

PR #502, opened 2022-02-01
fix display of permission comboboxes with scrollable columns across lines

fmigneault:
When permissions cannot fit in view, an horizontal scrollbar appears under the permission titles. 
Scrolling it moves all the selector boxes across all lines in the tree view accordingly.

## Changes

* Add UI horizontal multi-scroll (all tree lines simultaneously) of ``Permission`` selectors when there are more that
  can fit naturally within the tab view for a given ``Service`` type  (relates to #498).
* Widen permission "columns" for more permission titles to fit.
  The permission titles can still auto-scroll as well in cases were the permission name is just unreasonably long.
  The chosen dimension is the best trade-off of showing as must titles completely vs making the boxes too wide, needing extra scrolling to display all of them.

## Fixes

* minor fixes related to linting

### When there are too many permissions to display at once

image

### When space is sufficient

image

image

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -45,6 +45,9 @@ Features / Changes
 * Add auto-restore of previous scroll position in UI page following submitted form.
 * Add UI tooltip `Resource` ID to elements rendered in the ``Service`` and ``Permission`` hierarchy trees
   (relates to `#335 <https://github.com/Ouranosinc/Magpie/issues/335>`_).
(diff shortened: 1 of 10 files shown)


dbyrns (dismissed):
LGTM

f-PLT (approved):
Looks good to me too, and worked perfectly on my end.

PR #511, opened 2022-03-16
Permissions patch

ChaamC:
Content of this PR :

* Add ``PATCH /permissions`` endpoint that updates permissions and creates related resources if necessary.
* ``permissions.cfg`` now supports a new format for the ``type`` parameter, using multiple types separated
  by a slash character, matching each type with each resource found in the ``resource`` parameter.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,6 +9,12 @@ Changes
 `Unreleased <https://github.com/Ouranosinc/Magpie/tree/master>`_ (latest)
 ------------------------------------------------------------------------------------
 
+Features / Changes
+~~~~~~~~~~~~~~~~~~~~~
+* Add ``PATCH /permissions`` endpoint that updates permissions and creates related resources if necessary.
+* ``permissions.cfg`` now supports a new format for the ``type`` parameter, using multiple types separated
+  by a slash character, matching each type with each resource found in the ``resource`` parameter.
+
 Bug Fixes
 ~~~~~~~~~~~~~~~~~~~~~
 * Update linting configuration rules to validate all migration scripts employed by ``alembic``.
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,7 +12,7 @@ Changes
 Features / Changes
 ~~~~~~~~~~~~~~~~~~~~~
 * Add ``PATCH /permissions`` endpoint that updates permissions and creates related resources if necessary.
-* ``permissions.cfg`` now supports a new format for the ``type`` parameter, using multiple types separated
+* Add support of new format for ``permissions.cfg`` for the ``type`` parameter, using multiple types separated
   by a slash character, matching each type with each resource found in the ``resource`` parameter.
 
 .. _changes_3.23.0:
--- a/magpie/api/management/resource/resource_views.py
+++ b/magpie/api/management/resource/resource_views.py
@@ -12,6 +12,7 @@ from magpie.api.management.resource import resource_formats as rf
 from magpie.api.management.resource import resource_utils as ru
 from magpie.api.management.service.service_formats import format_service_resources
 from magpie.api.management.service.service_utils import get_services_by_type
+from magpie.api.management.user import user_utils as uu
 from magpie.permissions import PermissionType, format_permissions
 from magpie.register import magpie_register_permissions_from_config, sync_services_phoenix
 from magpie.services import SERVICE_TYPE_DICT, get_resource_child_allowed
@@ -171,23 +172,31 @@ def get_resource_types_view(request):
 @view_config(route_name=s.PermissionsAPI.name, request_method="PATCH")
 def update_permissions(request):
     """
-    Update the requested permissions.
+    Update the requested permissions and create missing related resources if necessary.
     """
     permissions = ar.get_value_multiformat_body_checked(request, "permissions", check_type=list)
-    ax.verify_param(permissions, not_none=True, not_empty=True,
-                    http_error=HTTPBadRequest, msg_on_fail=s.Permissions_PATCH_BadRequestResponseSchema.description)
+    ax.verify_param(permissions, not_none=True, not_empty=True, http_error=HTTPBadRequest,
+                    msg_on_fail="No permissions to update (empty `permissions` parameter.)")
 
     for entry in permissions:
-        # Check if user/group exists for each permission found
-        if "permission" in entry.keys() and entry["permission"]:
-            if "user" in entry.keys():
-                ax.verify_param(UserService.by_user_name(entry["user"], db_session=request.db),
-                                not_none=True, http_error=HTTPBadRequest,
-                                msg_on_fail=s.Permissions_PATCH_BadRequestResponseSchema.description)
-            if "group" in entry.keys():
+        ax.verify_param(entry, is_type=True, param_compare=dict, http_error=HTTPBadRequest,
+                        msg_on_fail="Permission entry should be of `dict` type, but type `{}` was found instead".format(
+                            type(entry)),
+                        content={"param_content": entry})
+        if "permission" in entry and entry["permission"]:
+            if "user" in entry:
+                user = UserService.by_user_name(entry["user"], db_session=request.db)
+                ax.verify_param(user, not_none=True, http_error=HTTPBadRequest,
+                                msg_on_fail="Permission's user `{}` could not be found in the database.".format(
+                                    entry["user"]),
+                                content={"param_content": entry})
+                uu.check_user_editable(user, request)
+            if "group" in entry:
                 ax.verify_param(GroupService.by_group_name(entry["group"], db_session=request.db),
                                 not_none=True, http_error=HTTPBadRequest,
-                                msg_on_fail=s.Permissions_PATCH_BadRequestResponseSchema.description)
+                                msg_on_fail="Permission's group `{}` could not be found in the database.".format(
+                                    entry["group"]),
+                                content={"param_content": entry})
 
     # Reformat permissions config
     permissions_cfg = {"permissions": []}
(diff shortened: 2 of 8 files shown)


magpie/register.py
```diff
@@ -712,7 +717,8 @@ def _parse_resource_path(permission_config_entry,   # type: PermissionConfigItem
                          entry_index,               # type: int
                          service_info,              # type: JSON
                          cookies_or_session=None,   # type: CookiesOrSessionType
-                         magpie_url=None,           # type: Optional[Str]
+                         magpie_url=None,           # type: Optional[Str],
+                         raise_errors=False         # type: bool
```
fmigneault, line 721: Following our discussion, `raise_errors` was needed because `_log_permission` didn't have to raise each time, but now it seems it is applied everywhere? 
If that is indeed the case, I prefer we go back to the original plan to have an `error_handler` function. 
I don't like that `_log_permission` raises an unrelated HTTP error for a call that is multiple layers above and in a completely different module.

magpie/register.py
```diff
@@ -898,7 +916,7 @@ def _apply_profile(_usr_name=None, _grp_name=None):
                 from magpie.api.management.group.group_utils import create_group
                 return create_group(**grp_data)
 
-    def _validate_response(operation, is_create, item_type="Permission"):
+    def _validate_response(operation, is_create, item_type="Permission", raise_errors=False):
```
fmigneault, line 919: Not needed to pass down the parameter since `_validate_response` is within `_apply_permission_entry`. 
Only parameters that change should be passed.

tests/interfaces.py
```diff
@@ -2384,6 +2384,176 @@ def test_DeleteUserServicePermission_ReservedKeyword_AnonymousForbidden(self):
...
+                    "group": self.test_group_name,
+                    "action": "create"
+                }
+            ]
+        }
+        resp = utils.test_request(self, "PATCH", "/permissions", headers=self.json_headers,
+                                  cookies=self.cookies, json=data)
+        utils.check_response_basic_info(resp, 200, expected_method="PATCH")
+
+        def check_permissions(data):
```
fmigneault, line 2419: Use a different parameter name for handling by the function.

magpie/api/schemas.py
```diff
@@ -777,6 +780,63 @@ class PermissionNameListSchema(colander.SequenceSchema):
...
+    permission = PermissionObjectSchema(
+        missing=colander.drop
+    )
+
+    action = colander.SchemaNode(
+        colander.String(),
+        description="Action to apply on the permission.",
+        example="create",
+        default="create",
+        validator=colander.OneOf(["create", "delete"])
```
fmigneault, line 809: supposed to be `remove`

fmigneault (changes requested):

magpie/register.py
ChaamC: `raise_errors` is not raised everywhere, well, the only exceptions are in another function :
https://github.com/Ouranosinc/Magpie/blob/48c2e6561afe55617c43a88561ba54199bc662da/magpie/register.py#L936-L941
https://github.com/Ouranosinc/Magpie/blob/48c2e6561afe55617c43a88561ba54199bc662da/magpie/register.py#L946-L951

Should I then still keep the `raise_errors` parameter?

magpie/register.py
```diff
@@ -739,6 +745,17 @@ def _parse_resource_path(permission_config_entry,   # type: PermissionConfigItem
         try:
             svc_name = service_info["service_name"]
             svc_type = service_info["service_type"]
+
+            # Prepare a list of types that fits with the list of resources
+            resource_type_list = resource_type_config.strip("/").split("/") if resource_type_config else [None]
```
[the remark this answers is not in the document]
ChaamC: Yes, I can apply it to the `resource_path`. Just to be sure though, there's a slight difference between the 2. Currently it was just removing the first and last character if it's a '/', and with the `strip` it will remove every prefix and suffix '/'. Wasn't sure if it was important for the `resource_path`.

magpie/register.py
fmigneault: What do you think of an optional `raise_error` function callable?
By default, `raise_error=None` would be replaced by a no-op `lambda msg, perm: None`. 
If `raise_error` is provided, it is not replaced and is used as is.
Then, everywhere that needs to raise would simply have `raise_error("msg", permission_config_entry)` after the `_log_permission`. 
That would give all the flexibility to the method calling `magpie_register_permissions_from_config` to specific how/which error to raise or not.

fmigneault: Just link this:
```python
_log_permission("Cannot generate resource", entry_index, detail="Service [{!s}] of type [{!s}] doesn't allows any sub-resource types. ".format(svc_name, svc_type))
raise_error("Cannot generate resource", permssion_config_entry)
continue  # in case 'raise_error' was no-op, to stop processing that perm-entry
```
Indeed that implies that we repeat the message... 
Maybe `_log_permission` (or another function) could be adjusted to instead return the formatted message, which would be passed to both functions?

ChaamC: If we do the msg formatting in another function, I would just remove the `_log_permission` function since it would just be a `LOGGER.log` call anyway. It would look something like :
```
msg = _format_log_permission("Cannot generate resource", entry_index, detail="Service [{!s}] of type [{!s}] doesn't allows any sub-resource types. ".format(svc_name, svc_type))
LOGGER.log(logging.WARN, msg)
raise_error(msg)
```

Are you sure we are not overthinking this though? This solution would require more repeated line of codes : we have to call the message formatter, the logger and the raise_error explicitly each time. In our initial solution, we just had to provide a boolean value to the `_log_permission` function and that's it. We had to pass that boolean value in all the related functions, but it is a bit similar in the other solution, where we have to pass the `raise_error` function callable in required functions.

fmigneault: I really don't like the fact that a function designed to log permission messages ends up raising an HTTP error. 
It is out of scope of what it is expected to do and also HTTP are involved only for API calls, which makes it out of scope for the entire register module. This is why I want an alternative where the initial call by the API is the one that provides the type of error (or if any) needs to be raised. 

I'm open to other ways to implement it though if you see an easier way, just brainstorming ideas.

fmigneault: An alternative would be to keep the original `raise_errors` (bool), and if enabled, raise `RegistrationConfigurationError` instead. The API call would have to catch it and re-raise it as HTTPBadRequest. 
For consistency, `_log_permission` should be renamed `_handle_permission`, where it does the log + optional raising.

ChaamC: You're right about the HTTP error, it does seem out of scope of the context of the `_log_permission`.

I think I like your last suggestion best. It combines the original solution, while rescoping things in a more coherent way, and the error logging will stay more concise (avoiding the formatting + logging + raising error calls each time.)

CHANGES.rst
```diff
@@ -9,7 +9,11 @@ Changes
 `Unreleased <https://github.com/Ouranosinc/Magpie/tree/master>`_ (latest)
 ------------------------------------------------------------------------------------
 
-* Nothing new for the moment.
+Features / Changes
+~~~~~~~~~~~~~~~~~~~~~
+* Add ``PATCH /permissions`` endpoint that updates permissions and creates related resources if necessary.
+* ``permissions.cfg`` now supports a new format for the ``type`` parameter, using multiple types separated
+  by a slash character, matching each type with each resource found in the ``resource`` parameter.
```
fmigneault, line 16: rewrite using "Add support of ..." format

magpie/api/management/resource/resource_views.py
```diff
@@ -212,3 +215,102 @@ def get_res_types(res):
...
+        ax.verify_param(user, not_none=True, http_error=HTTPBadRequest,
+                        msg_on_fail="Permission's user `{}` could not be found in the database.".format(user_name))
+        uu.check_user_editable(user, request)
+    for group_name in required_groups:
+        ax.verify_param(GroupService.by_group_name(group_name, db_session=request.db),
+                        not_none=True, http_error=HTTPBadRequest,
+                        msg_on_fail="Permission's group `{}` could not be found in the database.".format(group_name))
+
+    ax.verify_param(has_permission_to_update, is_true=True, http_error=HTTPBadRequest,
+                    msg_on_fail="No permissions to update (none of the input entries has a defined permission.)",
```
fmigneault, line 262: period after parenthesis

magpie/api/management/resource/resource_views.py
```diff
@@ -212,3 +215,102 @@ def get_res_types(res):
...
+                    msg_on_fail="No permissions to update (empty `permissions` parameter.)")
+
+    required_users = set()
+    required_groups = set()
+    has_permission_to_update = False
+    for entry in permissions:
+        ax.verify_param(entry, is_type=True, param_compare=dict, http_error=HTTPBadRequest,
+                        msg_on_fail="Permission entry should be of `dict` type, but type `{}` was found instead".format(
+                            type(entry)),
+                        content={"param_content": entry})
```
fmigneault, line 238: For this and following `verify_param`, use `param_content` (function parameter) with the entry provided as `{"value": entry}` instead of `content`. Using `content` overrides other error details added by the response handler.
You can also add the `param_name` argument for cases that help make it more explicit, like when `user`, `group`, `type`, etc. are checked. This will populate the `param` field in the response body.

magpie/register.py
```diff
@@ -703,6 +705,10 @@ def _log_permission(message, permission_index, trail=", skipping...", detail=Non
     permission = " [{!s}]".format(permission) if permission else ""
     LOGGER.log(level, "%s [permission #%d]%s%s", message, permission_index, permission, trail)
 
+    if raise_errors:
+        raise RegistrationConfigurationError(
+            "{} [permission #{}]{}{}".format(message, permission_index, permission, trail))
```
fmigneault, line 710: You can use the `msg = ...` with this format and pass it to both log/exception. 
No point to "optimize" the log call in this case.

tests/interfaces.py
```diff
@@ -2384,6 +2384,176 @@ def test_DeleteUserServicePermission_ReservedKeyword_AnonymousForbidden(self):
...
+                {
+                    "resource_name": "resource",
+                    "resource_type": "service"  # only the 1st resource should have the `service` type
+                }
+            ]
+        }
+        path = "/permissions"
+        resp = utils.test_request(self, "PATCH", path, headers=self.json_headers, expect_errors=True,
+                                  cookies=self.cookies, json=data)
+        utils.check_response_basic_info(resp, 400, expected_method="PATCH")
```
fmigneault, line 2555: For the 400 tests, run `check_error_param_structure` as well on the body (returned by this function).

fmigneault (changes requested):

fmigneault (approved):
Nice work.

PR #517, opened 2022-05-05
MagpieAdapter request/response hooks

fmigneault:
## Summary

The main purpose is presented here: https://github.com/Ouranosinc/Magpie/pull/517/files#diff-1c805585a74c7c16218b333063aed89c528c90d8ec5aa76257a1e248eda0a10fR139-R155
Using this config, pointing at file https://github.com/Ouranosinc/Magpie/blob/request-hooks/tests/hooks/request_hooks.py, `MagpieAdapter` would add the `X-WPS-Output-Context` header in the request sent during job creation toward Weaver using the requesting user ID, effectively producing the WPS-outputs in its user directory. 

## Changes

* Add JSON schema validation of loaded `Service` configuration (``providers.cfg``).
* Add optional ``hooks`` section under each `Service` definition of the ``providers.cfg`` or combined configuration
  file that allows pre/post request/response processing operations using plugin Python scripts.
* Store the validated `Service` configuration in ``magpie.services`` settings for later access to ``hooks`` definitions
  by the ``MagpieAdapter``.
* Rename the ``webhooks`` section stored in settings to ``magpie.webhooks`` to avoid possible name clashes.

## References

- Depends on bird-house/twitcher#114 to call the hooks. 
- Addresses DAC-145, DAC-168, DAC-467

## WIP

- [x] Add specific tests for `hooks`
- [x] Merge #516 (this PR is based on it) to reduce diff of actual changes by this PR

### DIFF
--- a/config/providers.cfg
+++ b/config/providers.cfg
@@ -147,19 +147,25 @@ providers:
     type: api
     sync_type: api
     hooks:
+      # NOTE: 'target' relative to MAGPIE_ROOT, not this file position
+      # below hooks are used for testing adapter functionalities
       - type: request
         path: "/processes/[w+_-]/jobs"
         method: POST
-        target: ../tests/hooks/request_hooks.py:add_x_wps_output_context
+        target: tests/hooks/request_hooks.py:add_x_wps_output_context
       - type: request
         path: "/jobs"
         method: POST
-        target: ../tests/hooks/request_hooks.py:add_x_wps_output_context
+        target: tests/hooks/request_hooks.py:add_x_wps_output_context
       - type: response
         path: "/processes/[w+_-]/jobs/[a-z0-9-]+"
         method: GET
-        target: ../tests/hooks/request_hooks.py:add_x_wps_output_link
+        target: tests/hooks/request_hooks.py:add_x_wps_output_link
       - type: response
         path: "/jobs/[a-z0-9-]+"
         method: GET
-        target: ../tests/hooks/request_hooks.py:add_x_wps_output_link
+        target: tests/hooks/request_hooks.py:add_x_wps_output_link
+      - type: response
+        path: "/jobs/[a-z0-9-]+"
+        method: GET
+        target: tests/hooks/request_hooks.py:combined_arguments
--- a/config/providers.cfg
+++ b/config/providers.cfg
@@ -20,7 +20,7 @@
 #               often equals to 'type' (see: magpie.cli.SYNC_SERVICES_TYPES)
 #   hooks:      list of request processing hooks for the service
 #
-# Hooks:
+# Hooks: (requires Magpie>=3.25.0, Twitcher>=0.7.0)
 # ------
 #   When items are specified and that the *original* request filters match a configuration, the processing hooks are
 #   applied onto the proxied request or response using it. Each hook definition must be provided using the following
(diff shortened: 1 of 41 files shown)


config/providers.cfg
```diff
@@ -108,3 +138,28 @@ providers:
...
+  weaver:
+    url: http://${HOSTNAME}:4001
+    title: weaver-ogc-api-processes
+    public: true
+    c4i: false
+    type: api
+    sync_type: api
+    hooks:
+      - type: request
+        path: "/processes/[w+_-]/jobs"
```
ChaamC, line 151: Does the regex segment only accept a single character here? (Same thing for other similar regex below)

magpie/adapter/__init__.py
```diff
@@ -243,3 +257,83 @@ def configurator_factory(self, container):  # noqa: R0201
...
+
+        .. versionadded:: 3.25
+            Requires ``Twitcher >= 0.7.x``.
+
+        The received response from the proxied service is normally returned directly.
+        This method can modify the response to adapt it for specific service logic.
+        """
+        response = self._apply_hooks(
+            response, service["name"], "response",
+            response.request.path, response.request.path, response.request.query_string
```
ChaamC, line 337: Should the first argument of this line be `response.request.method`?

ChaamC (commented):
Looks good, some minor comments to check, but that's it!

magpie/config.py
```diff
@@ -0,0 +1,96 @@
...
+SERVICE_HOOK_ITEM_SCHEMA = {
+    "type": "object",
+    "properties": {
+        "type": {"type": "string", "enum": ["request", "response"]},
+        "path": {"type": "string"},
+        "query": {"type": "string", "default": r".*"},
+        "method": {"type": "string", "default": "*", "enum": [
+            "HEAD", "GET", "POST", "PUT", "PATCH", "DELETE",
+            "head", "get", "post", "put", "patch", "delete",
+            "*"
```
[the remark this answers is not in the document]
fmigneault: Sadly we cannot since `AnyRequestMethod` exists only during `TYPE_CHECKING = 1` context when run by IDE/linters. 
We cannot do the other way around either, because `Literal` must be a list of literal strings (invalid to do something like `Literal[*list_values]`).

magpie/adapter/__init__.py
fmigneault: Indeed, was fixed in commit not yet pushed.

config/providers.cfg
fmigneault: Well seen. This regex is completely wrong. Fixed.

ChaamC (dismissed):

fmigneault:
@dbyrns PR is ready. I'm only working on the final test that works locally but fails in CI.

fmigneault:
Blocked by release pending https://github.com/bird-house/twitcher/pull/114#issuecomment-1129214580
CI test suite still only sees v0.6.2, so adapter-hook test operations are not executed.

dbyrns (approved):
Everything looks good in that implementation! Very versatile.
As discussed, to be 100% flexible, a new hook should be included in twitcher to literally allow to override the _send_request method which is not really generic. 
If you're worried about messing with the current implementation you could literally keep it as is in the default twitcher implementation, but allow MagpieAdapter to define it's own.

magpie/adapter/__init__.py
```diff
@@ -48,10 +55,17 @@
...
         warnings.warn(
             "Magpie version is not guaranteed to work with newer versions of Twitcher. "
-            "This Magpie version offers compatibility with Twitcher 0.6.x. "
+            "This Magpie version offers compatibility with Twitcher 0.6.x and 0.7.x."
+            "Current package versions are (Twitcher: {}, Magpie: {})".format(twitcher_version, magpie_version),
+            ImportWarning
+        )
+    elif LooseVersion(twitcher_version) < LooseVersion("0.7.0"):
+        warnings.warn(
+            "Magpie version offers more capabilities than Twitcher 0.6.x is able to provide. "
```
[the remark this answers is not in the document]
fmigneault: indeed

PR #518, opened 2022-05-20
dist to pypi as pavics-magpie

fmigneault:
PyPI requires that all `dist_require` packages are publicly accessible using `pip install <pkg>`. 
Because of this, URL commit/branch references are not permitted. 

The package distribution is blocked by the following PR (so new official versions are published) until Mapgie can be pushed as well:

- https://github.com/bbangert/beaker/pull/221
- https://github.com/authomatic/authomatic/pull/195

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -115,3 +115,26 @@ jobs:
           files: ./reports/coverage.xml
           fail_ci_if_error: true
           verbose: true
+
+  deploy_pypi:
+    needs: tests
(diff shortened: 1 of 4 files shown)


fmigneault:
@dbyrns Because there was a discussion about that recently, I gave a try to publish Magpie to PyPI. I found the issues listed in the PR description. Some of them make me feel it is not worth doing the change because it introduces a breaking change on existing platforms (mounted INI file in dockers).

dbyrns:
If there is no hard requirement for a pypi package, keeping this as a nice to have is ok.

dbyrns (approved):

PR #519, opened 2022-05-21
adapter request/response hook `context` parameter

fmigneault:
## Changes

Add ``context`` parameter available for `Service Hook` functions, offering more handlers to obtain references to
  the adapter, the request's `Service` implementation, and the corresponding database `Resource`. This can be used,
  amongst other things, to perform advanced operations such as validating other `Resource` and `Permission` conditions
  to modify the handled ``request`` or ``response`` by the hook.

## Context

When attempting to implement [DAC-178 *6.2.2 - Filter processes based on permissions*](https://crim-ca.atlassian.net/browse/DAC-178), it became quite obvious that the parameters offered by hooks were not sufficient. For filtering items returned by `/processes` for example, a check of all items in that response must be done. With the previous parameters, this involved recomputing many elements from the baser service JSON definitions, when in fact, many are already computed and fetched from the database for processing the request.

A single `context` parameter is used to "package" anything that could be necessary, mostly to limit a growing number of parameters over time. It also allows us to reuse some precomputed elements based on internal Magpie logic.

### DIFF
--- a/magpie/services.py
+++ b/magpie/services.py
@@ -243,13 +243,12 @@ class ServiceInterface(object):
         """
         Obtain the :term:`User` that was identified to obtain protected :term:`Resource` access.
         """
-        user = self.request.user
+        found = user = self.request.user
 
         # fallback lookup in case cookies exists but request method did not evaluate user
         # this can happen in situations where 'request.user' was pre-resolved when interacting with Twitcher
         if user is None:
             user = get_request_user(self.request)
-            self.request.user = user  # fix for future references using the expected location
 
         if not user:
             session = get_connected_session(self.request)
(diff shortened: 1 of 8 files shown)


magpie/services.py
```diff
@@ -243,20 +243,26 @@ def user_requested(self):
...
-            self.request.user = user  # fix for future references using the expected location
 
         if not user:
             session = get_connected_session(self.request)
             anonymous = get_constant("MAGPIE_ANONYMOUS_USER", self.request)
             user = UserService.by_user_name(anonymous, db_session=session)
             if user is None:
                 raise RuntimeError("No Anonymous user in the database")
+        # 'request.user' expected None (and should remain as such) when anonymous (unauthenticated)
+        elif found is None and user is not None:
```
dbyrns, line 260: The `and` second clause would always be true and thus useless since in the opposite the first `if` would kick in.

fmigneault: I put it explicitly because I found it was not obvious otherwise (depending on previous if compared to comment mentioning following elif).

dbyrns (dismissed):

ChaamC (approved):
LGTM, nothing to add, except maybe an open question. Should there be an update to the tests for the new `context` parameter?

PR #525, opened 2022-09-13
raise HTTP conflict on match permission update

fmigneault:
## Features / Changes

* Raise HTTP conflict on `Permission` update when the new `Permission` to be generated matches perfectly with the
  existing one to avoid unnecessary `Webhooks` calls (subsequent delete and create of the same `Permission` definition).

## Bug Fixes

* Fix ``verify_param`` indirectly causing an error with ``PermissionSet`` object passed as ``param_compare`` argument
  when attempting to resolve if that comparison argument is of ``str`` type, which was not handled as valid second
  argument for the ``PermissionSet.__eq__`` operation.

### DIFF
--- a/magpie/permissions.py
+++ b/magpie/permissions.py
@@ -152,10 +152,7 @@ class PermissionSet(object):
     def __eq__(self, other):
         # type: (Any) -> bool
         if not isinstance(other, PermissionSet):
-            if not (
-                isinstance(other, six.string_types) or
-                isinstance(other, (dict, tuple, PermissionTuple, Permission))
-            ):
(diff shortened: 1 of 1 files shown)


ChaamC (approved):
Should we add a test for the new case where we raise conflict for a permission exact match when applying a `PATCH /permissions` ? 

Approving this PR, since everything looks good. There's just this test question, if it is relevant.

fmigneault:
@ChaamC 
> Should we add a test for the new case where we raise conflict for a permission exact match when applying a `PATCH /permissions` ?

Based on test coverage results, it was already validated. I assume that it would behave the same whether called from PATCH request or another. I think the better test would be on cowbird side using this fix to validate how handlers behave in combination with many permission syncs and webhooks.

PR #582, opened 2023-05-30
add service type to permission webhooks

ChaamC:
Add the `service.type` data in the `Webhooks` for a `Permission` change, even if the resource is not a service,
since the root service type is necessary to `Cowbird`.

### DIFF
--- a/magpie/api/webhooks.py
+++ b/magpie/api/webhooks.py
@@ -14,6 +14,7 @@ from magpie.api import exception as ax
 from magpie.api import schemas as s
 from magpie.api.management.group.group_formats import format_group
 from magpie.api.management.resource.resource_formats import format_resource
+from magpie.api.management.resource import resource_utils as ru
 from magpie.api.management.service.service_formats import format_service
 from magpie.api.management.user.user_formats import format_user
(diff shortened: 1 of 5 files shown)


magpie/api/webhooks.py
```diff
@@ -166,7 +168,10 @@ def get_permission_update_params(target,         # type: Union[models.User, mode
     if resource.resource_type == "service":
         res_params = format_service(resource, basic_info=True, dotted=True)
     else:
-        res_params = {"service.{}".format(param): None for param in ["name", "type", "sync_type", "public_url"]}
+        res_params = {"service.{}".format(param): None for param in ["name", "sync_type", "public_url"]}
+        # Add the root service type even for non-service resources, since it is needed by Cowbird.
+        res_params["service.type"] = \
+            ru.get_resource_root_service_by_id(resource.resource_id, db_session=db_session).type
```
fmigneault, line 174: Use newline on parameters instead of the backslash.

fmigneault (commented):
Just minor formatting to adjust.

fmigneault (approved):

PR #589, opened 2023-09-14
network mode

mishaschwartz:
Introduce "Network Mode" which allows other Magpie instances to act as external authentication providers using JSON Web Tokens (JWT). 

This allows users registered across multiple Magpie instances in a network to more easily gain access to the resources within the network, without requiring the duplication of user credentials across the network.

In response to the discussion here: https://github.com/DACCS-Climate/DACCS-executive-committee/issues/8.

This change is fully backwards compatible and is only enabled if the `MAGPIE_NETWORK_MODE` is set (not set by default). Otherwise, magpie admins and regular users should see no changes at all.

TODO:

- [ ] add tests for new functionality
- [ ] add UI view to get a token (optional)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,7 +11,9 @@ Changes
 
 Features / Changes
 ~~~~~~~~~~~~~~~~~~~~~
-* n/a
+* Introduce "Network Mode" which allows other Magpie instances to act as external authentication providers using JSON
+  Web Tokens (JWT). This allows users registered across multiple Magpie instances in a network to more easily gain
+  access to the resources within the network, without requiring the duplication of user credentials across the network.
 
 Bug Fixes
 ~~~~~~~~~~~~~~~~~~~~~
--- a/docs/authentication.rst
+++ b/docs/authentication.rst
@@ -422,3 +422,84 @@ Furthermore, as described in the `procedure`_, :envvar:`MAGPIE_USER_REGISTRATION
 specify whether administrator approval is required or not. This additional step is purely up to the developers and
 server managers that use `Magpie` to decide if they desire more control over which individuals can join and access
 their services.
+
+.. _Network Mode:
+
+Network Mode
+------------
+
+If the :envvar:`MAGPIE_NETWORK_MODE` is enabled, an additional external authentication provider is added to Magpie which
+allows networked instances of Magpie to authenticate users for each other.
+
+Users can then log in to any Magpie instance where they have an account and request a personal network token in the form
+of a `_JSON Web Token <https://datatracker.ietf.org/doc/html/rfc7519>`_ which can be used to authenticate this user on
+any Magpie instance in the network.
+
+Managing the Network
+~~~~~~~~~~~~~~~~~~~~
+
+In order for Magpie instances to authenticate each other's users, each instance must be made aware of the existence of
+the others so that it knows where to send authentication verification requests.
+
+In order to register another Magpie instance as part of the same network, an admin user can create a
+:term:`Network Node` with a request to ``POST /network-nodes``. The parameters given to that request include a ``name``
+and a ``url``. The ``name`` is a the name of that other Magpie instance in the network and should correspond to the
+same value as the :envvar:`MAGPIE_INSTANCE_NAME` value set by the other Magpie instance. The ``url`` is a the root URL
+of the other Magpie instance.
+
+Once a :term:`Network Node` is registered, Magpie can use that other instance to authenticate users as long as the other
+instance also has :envvar:`MAGPIE_NETWORK_MODE` enabled.
+
+Managing Personal JWTs
+~~~~~~~~~~~~~~~~~~~~~~
+
+A :term:`User` can request a new network token with a request to the ``PATCH /token`` route. This route takes one
+optional parameter ``expires`` which is an integer indicating how long (in seconds) until that token expires, the
+default expiry for this token is :envvar:`MAGPIE_DEFAULT_TOKEN_EXPIRY`.
+
+Every time a :term:`User` makes a request to the ``PATCH /token`` route a new token is generated for them. This
+effectively cancels all previously created tokens for that user. If a user wishes to cancel all tokens, they can provide
+an ``expires`` value of ``0`` when making the request.
+
+Authentication
+~~~~~~~~~~~~~~
+
+Once a :term:`User` gets a personal network token, they can use that token to authenticate with any Magpie instance in
+the same network. When a user makes a request, they should set the ``provider_name`` parameter to the value of
+:envvar:`MAGPIE_NETWORK_PROVIDER` and provide the network token in the Authorization header in the following format:
+
+.. code-block:: http
+
+    Authorization: Bearer <network_token>
+
+When using the :ref:`Magpie Adapter <utilities_adapter>`, the token can also be passed as a parameter to the request,
+where the parameter name set by :envvar:`MAGPIE_NETWORK_TOKEN_NAME` and the value is the personal network token.
+
+Authorization
+~~~~~~~~~~~~~
+
+Managing authorization for :term:`Users` who authenticate using personal network tokens is complicated by the fact that
+a :term:`User` is not required to have a full account on both Magpie instances in order to using this authentication
+mechanism. This means that a :term:`User` may be logged in as a node-specific "anonymous" user.
+
+When another Magpie instance is registered as a :term:`Network Node`, a few additional entities are created:
+
+#. a group used to manage the permissions of all users who authenticate using the new :term:`Network Node`.
+   * this group's name will be the :envvar:`MAGPIE_NETWORK_NAME_PREFIX` followed by the :term:`Network Node` name
+#. a group used to manage the permissions of all users who authenticate using *any* other instance in the network
+   * this group's name will be the :envvar:`MAGPIE_NETWORK_GROUP_NAME`
+   * this group will only be created once, when the first :term:`Network Node` is registered
+#. an anonymous user that belongs to the two groups that were just created.
+   * this user name will be the :envvar:`MAGPIE_NETWORK_NAME_PREFIX` followed by the :term:`Network Node` name
+
+Here is an example to illustrate this point:
+
+* There are 3 Magpie instances in the network named A, B, and C
+* There is a :term:`User` named ``"toto"`` registered on instance A
+* There is no :term:`User` named ``"toto"`` who belongs to the ``"anonymous_network_A"`` group registered on instance B
+* There is a :term:`User` named ``"toto"`` who belongs to the ``"anonymous_network_A"`` group registered on instance C
+* Instance A is registered as a :term:`Network Node` on instances B and C
+* when ``"toto"`` gets a personal network token from instance A and uses it to log in on instance B they log in as the
+  the temporary ``"anonymous_network_A"`` user.
+* when ``"toto"`` gets a personal network token from instance A and uses it to log in on instance C they log in as the
+  ``"toto"`` user on instance C.
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -969,6 +969,83 @@ remain available as described at the start of the :ref:`Configuration` section.
     Name of the :term:`Provider` used for login. This represents the identifier that is set to define how to
     differentiate between a local sign-in procedure and a dispatched one some known :ref:`authn_providers`.
 
+Network Mode Settings
+~~~~~~~~~~~~~~~~~~~~~
+
+The following configuration parameters are related to Magpie's "Network Mode" which allows networked instances of Magpie
+to authenticate users for each other. All variables defined in this section are only used if
+:envvar:`MAGPIE_NETWORK_MODE` is enabled.
+
+.. envvar:: MAGPIE_NETWORK_MODE
+
+    [:class:`bool`]
+    (Default: ``False``)
+
+    .. versionadded:: 3.37
+
+    Enable "Network Mode" which enables all functionality to authenticate users using other Magpie instances as
+    external authentication providers.
+
+.. envvar:: MAGPIE_INSTANCE_NAME
+
+    [:class:`str`]
+
+    .. versionadded:: 3.37
+
+    The name of this Magpie instance in the network. This variable is used to determine if an authentication token was
+    issued by this instance of Magpie, or another instance in the network.
+
+    This variable is required if :envvar:`MAGPIE_NETWORK_MODE` is ``True``.
+
+.. envvar:: MAGPIE_DEFAULT_TOKEN_EXPIRY
+
+    [:class:`int`]
+    (Default: ``86400``)
+
+    .. versionadded:: 3.37
+
+    The default expiry time (in seconds) for an authentication token issued for the purpose of network authentication.
+
+.. envvar:: MAGPIE_NETWORK_TOKEN_NAME
+
+    [|constant|_]
+    (Value: ``"magpie_token"``)
+
+    .. versionadded:: 3.37
+
+    The name of the request parameter key whose value is the authentication token issued for the purpose of network
+    authentication.
+
+.. envvar:: MAGPIE_NETWORK_PROVIDER
+
+    [|constant|_]
+    (Value: ``"magpie_network"``)
+
+    .. versionadded:: 3.37
+
+    The name of the external provider that authenticates users using other Magpie instances as external authentication
+    providers.
+
+.. envvar:: MAGPIE_NETWORK_NAME_PREFIX
+
+    [|constant|_]
+    (Value: ``"anonymous_network_"``)
+
+    .. versionadded:: 3.37
+
+    A prefix added to the anonymous network user and network group names. These names are constructed by prepending the
+    remote Magpie instance name with this prefix. For example, a Magpie instance named ``"example123"`` will have a
+    corresponding user and group named ``"anonymous_network_example123"``.
+
+.. envvar:: MAGPIE_NETWORK_GROUP_NAME
+
+    [|constant|_]
+    (Value: ``"magpie_network"``)
+
+    .. versionadded:: 3.37
+
+    The name of the group created to manage permissions for all users authenticated using Magpie instances as external
+    authentication providers.
 
 .. _config_phoenix:
 
--- a/docs/glossary.rst
+++ b/docs/glossary.rst
@@ -138,6 +138,16 @@ Glossary
         :py:data:`magpie.constants.MAGPIE_ANONYMOUS_USER`. Otherwise, it is whoever the
         :term:`Authentication` mechanism identifies with token extracted from request :term:`Cookies`.
 
+    Network Node
+        A reference to an instance of the Magpie software within a network of Magpie instances. Each Magpie instance
+        within the network is registered in the database as a row in the ``network_nodes`` table. Each node is
+        represented by a name that is unique across all nodes in the network, and a url that is used to send http
+        requests to that specific node.
+
+    Network Token
+        A unique random string that can be used to authenticate a user as part of the :ref:`Network Mode` authentication
+        procedure.
+
     OpenAPI
     OAS
         The |OpenAPI-spec|_ (`OAS`) defines a standard, programming language-agnostic interface description for
--- a/magpie/adapter/magpieowssecurity.py
+++ b/magpie/adapter/magpieowssecurity.py
@@ -269,11 +269,18 @@ class MagpieOWSSecurity(OWSSecurityInterface):
         """
         settings = get_settings(request)
         token_name = get_constant("MAGPIE_COOKIE_NAME", settings_container=settings)
+        network_mode = get_constant("MAGPIE_NETWORK_MODE", settings_container=settings,
+                                    settings_name="magpie.network_mode")
+        headers = dict(request.headers)
+        network_token_name = get_constant("MAGPIE_NETWORK_TOKEN_NAME", settings_container=settings)
+        if network_mode and "Authorization" not in headers and network_token_name in request.params:
+            headers["Authorization"] = "Bearer {}".format(request.params[network_token_name])
+            request.params["provider_name"] = request.params.get("provider_name",
+                                                                 get_constant("MAGPIE_NETWORK_PROVIDER", settings))
         if "Authorization" in request.headers and token_name not in request.cookies:
             magpie_prov = request.params.get("provider_name", get_constant("MAGPIE_DEFAULT_PROVIDER", settings))
             magpie_path = ProviderSigninAPI.path.format(provider_name=magpie_prov)
             magpie_auth = "{}{}".format(self.magpie_url, magpie_path)
-            headers = dict(request.headers)
             headers.update({"Homepage-Route": "/session", "Accept": CONTENT_TYPE_JSON})
             session_resp = requests.get(magpie_auth, headers=headers, verify=self.twitcher_ssl_verify)
             if session_resp.status_code != HTTPOk.code:
--- a/magpie/alembic/versions/2023-08-25_2cfe144538e8_add_network_tables.py
+++ b/magpie/alembic/versions/2023-08-25_2cfe144538e8_add_network_tables.py
@@ -1,0 +1,54 @@
+"""
+Add Network_Tokens Table
+
+Revision ID: 2cfe144538e8
+Revises: 5e5acc33adce
+Create Date: 2023-08-25 13:36:16.930374
+"""
+import uuid
+
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.dialects.postgresql import UUID
+from sqlalchemy.orm.session import sessionmaker
+from sqlalchemy_utils import URLType
+
+# Revision identifiers, used by Alembic.
+# pylint: disable=C0103,invalid-name  # revision control variables not uppercase
+revision = "2cfe144538e8"
+down_revision = "5e5acc33adce"
+branch_labels = None
+depends_on = None
+
+Session = sessionmaker()
+
+
+def upgrade():
+    op.create_table("network_tokens",
+                    sa.Column("token", UUID(as_uuid=True),
+                              primary_key=True, default=uuid.uuid4, unique=True),
+                    sa.Column("user_id", sa.Integer,
+                              sa.ForeignKey("users.id", onupdate="CASCADE", ondelete="CASCADE"), unique=True)
+                    )
+    op.create_table("network_nodes",
+                    sa.Column("id", sa.Integer, primary_key=True, autoincrement=True),
+                    sa.Column("name", sa.Unicode(128), nullable=False, unique=True),
+                    sa.Column("url", URLType(), nullable=False)
+                    )
+    op.add_column("users", sa.Column("network_node_id", sa.Integer,
+                                     sa.ForeignKey("network_nodes.id", onupdate="CASCADE", ondelete="CASCADE"),
+                                     nullable=True))
+    op.drop_constraint("uq_users_user_name", "users")
+    op.create_unique_constraint("uq_users_user_name_network_node_id", "users", ["user_name", "network_node_id"])
+    op.drop_constraint("uq_users_email", "users")
+    op.create_unique_constraint("uq_users_email_network_node_id", "users", ["email", "network_node_id"])
+
+
+def downgrade():
+    op.drop_constraint("uq_users_user_name_network_node_id", "users")
+    op.create_unique_constraint("uq_users_user_name", "users", ["user_name"])
+    op.drop_constraint("uq_users_email_network_node_id", "users")
+    op.create_unique_constraint("uq_users_email", "users", ["email"])
+    op.drop_table("network_tokens")
+    op.drop_table("network_nodes")
+    op.drop_column("users", "network_node_id")
--- a/magpie/api/login/login.py
+++ b/magpie/api/login/login.py
@@ -1,6 +1,10 @@
 import json
+import uuid
+from datetime import datetime, timedelta
 from typing import TYPE_CHECKING
 
+import jwt
+import requests
 from authomatic.adapters import WebObAdapter
 from authomatic.core import Credentials, LoginResult, resolve_provider_class
 from authomatic.exceptions import OAuth2Error
@@ -34,7 +38,7 @@ from magpie.api import requests as ar
 from magpie.api import schemas as s
 from magpie.api.management.user.user_formats import format_user
 from magpie.api.management.user.user_utils import create_user
-from magpie.constants import get_constant
+from magpie.constants import get_constant, protected_user_name_regex
 from magpie.security import authomatic_setup, get_providers
 from magpie.utils import (
     CONTENT_TYPE_JSON,
@@ -47,14 +51,23 @@ from magpie.utils import (
 
 if TYPE_CHECKING:
     from magpie.typedefs import Session, Str
+    from typing import Optional
 
 LOGGER = get_logger(__name__)
 
+MAGPIE_NETWORK_MODE = get_constant("MAGPIE_NETWORK_MODE", settings_name="magpie.network_mode")
 
 # dictionaries of {'provider_id': 'provider_display_name'}
 MAGPIE_DEFAULT_PROVIDER = get_constant("MAGPIE_DEFAULT_PROVIDER")
 MAGPIE_INTERNAL_PROVIDERS = {MAGPIE_DEFAULT_PROVIDER: MAGPIE_DEFAULT_PROVIDER.capitalize()}
 MAGPIE_EXTERNAL_PROVIDERS = get_providers()
+
+if MAGPIE_NETWORK_MODE:
+    MAGPIE_NETWORK_TOKEN_NAME = get_constant("MAGPIE_NETWORK_TOKEN_NAME")
+    MAGPIE_NETWORK_PROVIDER = get_constant("MAGPIE_NETWORK_PROVIDER")
+    MAGPIE_INSTANCE_NAME = get_constant("MAGPIE_INSTANCE_NAME")
+    MAGPIE_EXTERNAL_PROVIDERS[MAGPIE_NETWORK_PROVIDER] = MAGPIE_NETWORK_PROVIDER
+
 MAGPIE_PROVIDER_KEYS = frozenset(set(MAGPIE_INTERNAL_PROVIDERS) | set(MAGPIE_EXTERNAL_PROVIDERS))
 
 
@@ -130,8 +143,8 @@ def sign_in_view(request):
     # request handler. Catch that specific exception and return it to bypass the EXCVIEW tween that result in that
     # automatic convert to return 403 directly.
     try:
-        anonymous = get_constant("MAGPIE_ANONYMOUS_USER", request)
-        ax.verify_param(user_name, not_equal=True, param_compare=anonymous, param_name="user_name",
+        anonymous_regex = protected_user_name_regex(include_admin=False, settings_container=request)
+        ax.verify_param(user_name, not_matches=True, param_compare=anonymous_regex, param_name="user_name",
                         http_error=HTTPForbidden, msg_on_fail=s.Signin_POST_ForbiddenResponseSchema.description)
     except HTTPForbidden as http_error:
         return http_error
@@ -239,17 +252,12 @@ def new_user_external(external_user_name, external_id, email, provider_name, db_
     return user
 
 
-def login_success_external(request, external_user_name, external_id, email, provider_name):
-    # type: (Request, Str, Str, Str, Str) -> HTTPException
+def login_success_external(request, user):
+    # type: (Request, models.User) -> HTTPException
     """
     Generates the login response in case of successful external provider identification.
     """
-    # find possibly already registered user by external_id/provider
-    user = ExternalIdentityService.user_by_external_id_and_provider(external_id, provider_name, request.db)
-    if user is None:
-        # create new user with an External Identity
-        user = new_user_external(external_user_name=external_user_name, external_id=external_id,
-                                 email=email, provider_name=provider_name, db_session=request.db)
+
     # set a header to remember user (set-cookie)
     headers = remember(request, user.id)
 
@@ -271,10 +279,27 @@ def login_success_external(request, external_user_name, external_id, email, prov
                          http_kwargs={"location": homepage_route, "headers": headers})
 
 
+def user_from_token(request, token):
+    # type: (Request, Str) -> Optional[models.User]
+    """
+    Return the ``User`` associated with the token as long as the token is valid and issued by this Magpie instance.
+    """
+    magpie_secret = get_constant("MAGPIE_SECRET", request, settings_name="magpie.secret")
+    decoded_token = jwt.decode(token, magpie_secret, algorithms="HS256", issuer=MAGPIE_INSTANCE_NAME,
+                               options={"require": ["exp", "token"]})
+
+    network_token = (request.db.query(models.NetworkToken)
+                     .join(models.User)
+                     .filter(models.NetworkToken.token == decoded_token["token"])
+                     .first())
+    if network_token:
+        return network_token.user
+
+
 @s.ProviderSigninAPI.get(schema=s.ProviderSignin_GET_RequestSchema, tags=[s.SessionTag],
                          response_schemas=s.ProviderSignin_GET_responses)
 @view_config(route_name=s.ProviderSigninAPI.name, permission=NO_PERMISSION_REQUIRED)
-def authomatic_login_view(request):
+def external_login_view(request):
     """
     Signs in a user session using an external provider.
     """
@@ -282,58 +307,99 @@ def authomatic_login_view(request):
     response = Response()
     verify_provider(provider_name)
     try:
-        authomatic_handler = authomatic_setup(request)
+        if provider_name == MAGPIE_NETWORK_PROVIDER:
+            if "Authorization" in request.headers:
+                token_type, token = request.headers.get("Authorization").split()
+                if token_type != "Bearer":
+                    ax.raise_http(http_error=HTTPBadRequest,
+                                  detail=s.ProviderSignin_GET_BadRequestResponseSchema.description)
+                authenticated_user = None
+                try:
+                    authenticated_user = user_from_token(request, token)
+                except jwt.exceptions.InvalidIssuerError:
+                    decoded_token = jwt.decode(token, options={"verify_signature": False})
+                    issuer = decoded_token.get("iss")
+                    node = request.db.query(models.NetworkNode).filter(models.NetworkNode.name == issuer).first()
+                    if node:
+                        response = requests.get("{}{}".format(node.url, s.TokenValidateAPI.path),
+                                                headers={"Accept": CONTENT_TYPE_JSON})
 
-        # if we directly have the Authorization header, bypass authomatic login and retrieve 'userinfo' to signin
-        if "Authorization" in request.headers and "authomatic" not in request.cookies:
-            provider_config = authomatic_handler.config.get(provider_name, {})
-            provider_class = resolve_provider_class(provider_config.get("class_"))
-            provider = provider_class(authomatic_handler, adapter=None, provider_name=provider_name)
-            # provide the token user data, let the external provider update it on login afterwards
-            token_type, access_token = request.headers.get("Authorization").split()
-            data = {"access_token": access_token, "token_type": token_type}
-            cred = Credentials(authomatic_handler.config, token=access_token, token_type=token_type, provider=provider)
-            provider.credentials = cred
-            result = LoginResult(provider)
-            # pylint: disable=W0212
-            result.provider.user = result.provider._update_or_create_user(data, credentials=cred)  # noqa: W0212
-
-        # otherwise, use the standard login procedure
-        else:
-            result = authomatic_handler.login(WebObAdapter(request, response), provider_name)
-            if result is None:
-                if response.location is not None:
-                    return HTTPTemporaryRedirect(location=response.location, headers=response.headers)
-                return response
-
-        if result:
-            if result.error:
-                # Login procedure finished with an error.
-                error = result.error.to_dict() if hasattr(result.error, "to_dict") else result.error
-                LOGGER.debug("Login failure with error. [%r]", error)
-                return login_failure_view(request, reason=result.error.message)
-            if result.user:
-                # OAuth 2.0 and OAuth 1.0a provide only limited user data on login,
-                # update the user to get more info.
-                if not (result.user.name and result.user.id):
-                    try:
-                        response = result.user.update()
-                    # this error can happen if providing incorrectly formed authorization header
-                    except OAuth2Error as exc:
-                        LOGGER.debug("Login failure with Authorization header.")
-                        ax.raise_http(http_error=HTTPBadRequest, content={"reason": str(exc.message)},
-                                      detail=s.ProviderSignin_GET_BadRequestResponseSchema.description)
-                    # verify that the update procedure succeeded with provided token
-                    if 400 <= response.status < 500:
-                        LOGGER.debug("Login failure with invalid token.")
+                        if response.status_code != HTTPOk.code:
+                            ax.raise_http(http_error=HTTPUnauthorized,
+                                          detail=s.ProviderSignin_GET_UnauthorizedTokenSchema.description)
+                        user_name = response.json().get("user_name")
+                        authenticated_user = (node.users.filter(models.User.name == user_name).first() or
+                                              node.anonymous_user)
+                    else:
                         ax.raise_http(http_error=HTTPUnauthorized,
-                                      detail=s.ProviderSignin_GET_UnauthorizedResponseSchema.description)
-                # create/retrieve the user using found details from login provider
-                return login_success_external(request,
-                                              external_id=result.user.username or result.user.id,
-                                              email=result.user.email,
-                                              provider_name=result.provider.name,
-                                              external_user_name=result.user.name)
+                                      detail=s.ProviderSignin_GET_UnauthorizedTokenSchema.description)
+                except jwt.exceptions.PyJWTError as exc:
+                    ax.raise_http(http_error=HTTPUnauthorized, content={"reason": str(exc)},
+                                  detail=s.ProviderSignin_GET_UnauthorizedTokenSchema.description)
+                if authenticated_user:
+                    return login_success_external(request, authenticated_user)
+            else:
+                ax.raise_http(http_error=HTTPBadRequest,
+                              detail=s.ProviderSignin_GET_BadRequestResponseSchema.description)
+        else:
+            authomatic_handler = authomatic_setup(request)
+
+            # if we directly have the Authorization header, bypass authomatic login and retrieve 'userinfo' to signin
+            if "Authorization" in request.headers and "authomatic" not in request.cookies:
+                provider_config = authomatic_handler.config.get(provider_name, {})
+                provider_class = resolve_provider_class(provider_config.get("class_"))
+                provider = provider_class(authomatic_handler, adapter=None, provider_name=provider_name)
+                # provide the token user data, let the external provider update it on login afterwards
+                token_type, access_token = request.headers.get("Authorization").split()
+                data = {"access_token": access_token, "token_type": token_type}
+                cred = Credentials(authomatic_handler.config, token=access_token, token_type=token_type,
+                                   provider=provider)
+                provider.credentials = cred
+                result = LoginResult(provider)
+                # pylint: disable=W0212
+                result.provider.user = result.provider._update_or_create_user(data, credentials=cred)  # noqa: W0212
+
+            # otherwise, use the standard login procedure
+            else:
+                result = authomatic_handler.login(WebObAdapter(request, response), provider_name)
+                if result is None:
+                    if response.location is not None:
+                        return HTTPTemporaryRedirect(location=response.location, headers=response.headers)
+                    return response
+
+            if result:
+                if result.error:
+                    # Login procedure finished with an error.
+                    error = result.error.to_dict() if hasattr(result.error, "to_dict") else result.error
+                    LOGGER.debug("Login failure with error. [%r]", error)
+                    return login_failure_view(request, reason=result.error.message)
+                if result.user:
+                    # OAuth 2.0 and OAuth 1.0a provide only limited user data on login,
+                    # update the user to get more info.
+                    if not (result.user.name and result.user.id):
+                        try:
+                            response = result.user.update()
+                        # this error can happen if providing incorrectly formed authorization header
+                        except OAuth2Error as exc:
+                            LOGGER.debug("Login failure with Authorization header.")
+                            ax.raise_http(http_error=HTTPBadRequest, content={"reason": str(exc.message)},
+                                          detail=s.ProviderSignin_GET_BadRequestResponseSchema.description)
+                        # verify that the update procedure succeeded with provided token
+                        if 400 <= response.status < 500:
+                            LOGGER.debug("Login failure with invalid token.")
+                            ax.raise_http(http_error=HTTPUnauthorized,
+                                          detail=s.ProviderSignin_GET_UnauthorizedResponseSchema.description)
+                    # create/retrieve the user using found details from login provider
+                    # find possibly already registered user by external_id/provider
+                    external_id = result.user.username or result.user.id
+                    user = ExternalIdentityService.user_by_external_id_and_provider(external_id, provider_name,
+                                                                                    request.db)
+                    if user is None:
+                        # create new user with an External Identity
+                        user = new_user_external(external_user_name=result.user.name, external_id=external_id,
+                                                 email=result.user.email, provider_name=result.provider.name,
+                                                 db_session=request.db)
+                    return login_success_external(request, user)
     except Exception as exc:
         exc_msg = "Unhandled error during external provider '{}' login. [{!s}]".format(provider_name, exc)
         LOGGER.exception(exc_msg, exc_info=True)
@@ -371,6 +437,71 @@ def get_session_view(request):
     session_json = ax.evaluate_call(lambda: _get_session(request), http_error=HTTPInternalServerError,
                                     msg_on_fail=s.Session_GET_InternalServerErrorResponseSchema.description)
     return ax.valid_http(http_success=HTTPOk, detail=s.Session_GET_OkResponseSchema.description, content=session_json)
+
+
+if MAGPIE_NETWORK_MODE:
+    @s.TokenAPI.patch(schema=s.Token_PATCH_RequestBodySchema, tags=[s.SessionTag],
+                      response_schemas=s.Token_PATCH_responses)
+    @view_config(route_name=s.TokenAPI.name, request_method="PATCH", permission=Authenticated)
+    def token_view(request):
+        """
+        Get a token. Generates a new token every time this route is accessed so this can also
+        be used to invalidate a previously generated token.
+        """
+        magpie_secret = get_constant("MAGPIE_SECRET", request, settings_name="magpie.secret")
+        default_expiry = get_constant("MAGPIE_DEFAULT_TOKEN_EXPIRY", request,
+                                      settings_name="magpie.default_token_expiry")
+        magpie_instance_name = get_constant("MAGPIE_INSTANCE_NAME",
+                                            request,
+                                            settings_name="magpie.instance_name")
+
+        expires = request.GET.get("expires", default_expiry)
+        try:
+            expires = int(expires)
+        except ValueError:
+            expires = int(default_expiry)
+
+        expiry = datetime.utcnow() + timedelta(seconds=expires)
+
+        def upsert_token():
+            network_token = request.db.query(models.NetworkToken).filter(
+                models.NetworkToken.user_id == request.user.id).first()
+            if network_token:
+                network_token.token = uuid.uuid4()
+            else:
+                network_token = models.NetworkToken(user_id=request.user.id)
+                request.db.add(network_token)
+            return network_token.token
+
+        token_value = ax.evaluate_call(lambda: upsert_token(), fallback=lambda: request.db.rollback(),
+                                       http_error=HTTPInternalServerError)
+
+        token = jwt.encode({"token": str(token_value),
+                            "exp": expiry,
+                            "iss": magpie_instance_name}, magpie_secret, algorithm="HS256")
+
+        return ax.valid_http(http_success=HTTPOk, detail=s.Token_PATCH_OkResponseSchema.description,
+                             content={"token": token})
+
+
+    @s.TokenValidateAPI.get(schema=s.TokenValidate_GET_RequestBodySchema, tags=[s.SessionTag],
+                            response_schemas=s.TokenValidate_GET_responses)
+    @view_config(route_name=s.TokenValidateAPI.name, permission=NO_PERMISSION_REQUIRED)
+    def validate_token_view(request):
+        """ Validate a token """
+        token = request.GET.get("token")
+        try:
+            authenticated_user = user_from_token(request, token)
+            if authenticated_user:
+                return ax.valid_http(http_success=HTTPOk,
+                                     detail=s.TokenValidate_GET_OkResponseSchema.description,
+                                     content={"user_name": authenticated_user.user_name})
+            else:
+                ax.raise_http(http_error=HTTPUnauthorized,
+                              detail=s.TokenValidate_GET_BadRequestResponseSchema.description)
+        except jwt.exceptions.PyJWTError as exc:
+            ax.raise_http(http_error=HTTPUnauthorized, content={"reason": str(exc)},
+                          detail=s.TokenValidate_GET_BadRequestResponseSchema.description)
 
 
 @s.ProvidersAPI.get(tags=[s.SessionTag], response_schemas=s.Providers_GET_responses)
--- a/magpie/api/management/network_node/network_node_views.py
+++ b/magpie/api/management/network_node/network_node_views.py
@@ -1,0 +1,89 @@
+from pyramid.httpexceptions import (
+    HTTPBadRequest,
+    HTTPConflict,
+    HTTPNotFound,
+    HTTPOk,
+    HTTPCreated,
+    HTTPInternalServerError
+)
+
+from pyramid.view import view_config
+
+from magpie import models
+from magpie.api import exception as ax
+from magpie.api import requests as ar
+from magpie.api import schemas as s
+from magpie.api.management.network_node.network_node_utils import create_network_node, delete_network_node
+from magpie.constants import get_constant
+
+
+if get_constant("MAGPIE_NETWORK_MODE", settings_name="magpie.network_mode"):
+    @s.NetworkNodesAPI.get(tags=[s.NetworkNodeTag], response_schemas=s.NetworkNodes_GET_responses)
+    @view_config(route_name=s.NetworkNodesAPI.name, request_method="GET")
+    def get_network_nodes_view(request):
+        nodes = [{"name": n.name, "url": n.url} for n in request.db.query(models.NetworkNode).all()]
+        return ax.valid_http(http_success=HTTPOk, detail=s.NetworkNode_GET_OkResponseSchema.description,
+                             content={"nodes": nodes})
+
+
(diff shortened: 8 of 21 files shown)


magpie/api/management/network_node/network_node_views.py
```diff
@@ -0,0 +1,89 @@
...
+                             content={"name": node.name, "url": node.url})
+
+
+    @s.NetworkNodesAPI.post(schema=s.NetworkNode_POST_RequestBodySchema, tags=[s.NetworkNodeTag],
+                            response_schemas=s.NetworkNodes_POST_responses)
+    @view_config(route_name=s.NetworkNodesAPI.name, request_method="POST")
+    def create_network_node_view(request):
+        node_name = request.POST.get("name")
+        node_url = request.POST.get("url")
+        ax.verify_param(all([node_name, node_url]), is_true=True,
```
fmigneault, line 47: Should use the checking utilities to validate each parameter against some regex. This only ensures there is a non-empty string, but it could be any value that could cause issues if they are malformed.

magpie/api/login/login.py
```diff
@@ -271,69 +279,127 @@ def login_success_external(request, external_user_name, external_id, email, prov
...
+        if provider_name == MAGPIE_NETWORK_PROVIDER:
+            if "Authorization" in request.headers:
+                token_type, token = request.headers.get("Authorization").split()
+                if token_type != "Bearer":
+                    ax.raise_http(http_error=HTTPBadRequest,
+                                  detail=s.ProviderSignin_GET_BadRequestResponseSchema.description)
+                authenticated_user = None
+                try:
+                    authenticated_user = user_from_token(request, token)
+                except jwt.exceptions.InvalidIssuerError:
```
fmigneault, line 319: Perform error handling closer to the part where it actually happens. Return `None` and handle accordingly here.

magpie/api/management/network_node/network_node_views.py
```diff
@@ -0,0 +1,89 @@
...
+    def update_network_node_view(request):
+        node_name = ar.get_value_matchdict_checked(request, "node")
+        node = ax.evaluate_call(
+            lambda: request.db.query(models.NetworkNode).filter(models.NetworkNode.name == node_name).one(),
+            http_error=HTTPNotFound,
+            msg_on_fail=s.NetworkNode_GET_NotFoundResponseSchema.description)
+        new_name = request.POST.get("name")
+        new_url = request.POST.get("url")
+        ax.verify_param(any([new_name, new_url]), is_true=True,
+                        http_error=HTTPBadRequest, msg_on_fail=s.BadRequestResponseSchema.description)
```
fmigneault, line 67: Same checks.

magpie/api/management/network_node/network_node_views.py
```diff
@@ -0,0 +1,89 @@
...
+    @view_config(route_name=s.NetworkNodesAPI.name, request_method="POST")
+    def create_network_node_view(request):
+        node_name = request.POST.get("name")
+        node_url = request.POST.get("url")
+        ax.verify_param(all([node_name, node_url]), is_true=True,
+                        http_error=HTTPBadRequest, msg_on_fail=s.BadRequestResponseSchema.description)
+        ax.evaluate_call(lambda: create_network_node(request, node_name, node_url),
+                         http_error=HTTPConflict,
+                         msg_on_fail=s.NetworkNodes_POST_ConflictResponseSchema.description)
+        return ax.valid_http(http_success=HTTPCreated, detail=s.NetworkNode_GET_OkResponseSchema.description)
```
fmigneault, line 52: Might need more than this. 

For now, any admin could register a node, which could be updated/deleted/recreated at some other point by any other admin, without any handshake validation of the node and associated tokens generated for it. If one node in the network gets hacked, that could allow a malicious user to generate tokens over the entire network.

Similar to when registering an external auth provider, there should be some kind of client ID/secret pair that is generated, and which the user must register back on their side to sign the tokens for that new node. This way, when a token is generated for a given node and sent to it, that node would have some ID to validate the token against and ensure it comes from a trusted location.

fmigneault (changes requested):
Not finished full review. I will need to study the features and security aspects more. Could need to discuss about the feature design.

I'm not sure if I'm sold on the `MAGPIE_NETWORK_NAME_PREFIX` approach to generate new users/groups. This can cause weird situations, such as permissions being set on two entities that should represent the same user. It would be better to have a distinct table, such as in the case of pending users and external identities, and make a link to users to "combine" their various network references. 

It seems like the custom external login code is essentially doing what `authomatic` would perform, but using non-standard OAuth methodology and endpoints.

magpie/alembic/versions/2023-08-25_2cfe144538e8_add_network_tables.py
```diff
@@ -0,0 +1,54 @@
...
+                    )
+    op.create_table("network_nodes",
+                    sa.Column("id", sa.Integer, primary_key=True, autoincrement=True),
+                    sa.Column("name", sa.Unicode(128), nullable=False, unique=True),
+                    sa.Column("url", URLType(), nullable=False)
+                    )
+    op.add_column("users", sa.Column("network_node_id", sa.Integer,
+                                     sa.ForeignKey("network_nodes.id", onupdate="CASCADE", ondelete="CASCADE"),
+                                     nullable=True))
+    op.drop_constraint("uq_users_user_name", "users")
```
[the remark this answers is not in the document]
mishaschwartz: Yeah I tried to search through to catch all the cases for the current searches and enforce that the associated network node be null. Another pair of eyes would be great.

magpie/api/schemas.py
```diff
@@ -435,6 +461,9 @@ class TemporaryURL_RequestPathSchema(colander.MappingSchema):
     ServicesTag: "Management of service definitions, children resources and their applicable permissions.",
 }
 
+if get_constant("MAGPIE_NETWORK_MODE", settings_name="magpie.network_mode"):
+    TAG_DESCRIPTIONS[NetworkNodeTag] = "Management of references to other Magpie instances in the network."
```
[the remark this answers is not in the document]
mishaschwartz: Maybe I don't understand the order here. When does the INI file get parsed? At run time, after the python files are parsed but before the server starts?

docs/configuration.rst
```diff
@@ -969,6 +969,83 @@ remain available as described at the start of the :ref:`Configuration` section.
...
+The following configuration parameters are related to Magpie's "Network Mode" which allows networked instances of Magpie
+to authenticate users for each other. All variables defined in this section are only used if
+:envvar:`MAGPIE_NETWORK_MODE` is enabled.
+
+.. envvar:: MAGPIE_NETWORK_MODE
+
+    [:class:`bool`]
+    (Default: ``False``)
+
+    .. versionadded:: 3.37
```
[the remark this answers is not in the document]
mishaschwartz: I think I assumed that this change would be its own version and the current unreleased changes would be made into version 3.36 before then. But I don't know how you decide on your releases for Magpie so I can change this if that's not the case.

magpie/api/login/login.py
```diff
@@ -373,6 +439,71 @@ def _get_session(req):
...
+        be used to invalidate a previously generated token.
+        """
+        magpie_secret = get_constant("MAGPIE_SECRET", request, settings_name="magpie.secret")
+        default_expiry = get_constant("MAGPIE_DEFAULT_TOKEN_EXPIRY", request,
+                                      settings_name="magpie.default_token_expiry")
+        magpie_instance_name = get_constant("MAGPIE_INSTANCE_NAME",
+                                            request,
+                                            settings_name="magpie.instance_name")
+
+        expires = request.GET.get("expires", default_expiry)
```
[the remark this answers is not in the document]
mishaschwartz: No it's an error. Thanks for the catch

docs/authentication.rst
```diff
@@ -422,3 +422,84 @@ Furthermore, as described in the `procedure`_, :envvar:`MAGPIE_USER_REGISTRATION
...
+
+* There are 3 Magpie instances in the network named A, B, and C
+* There is a :term:`User` named ``"toto"`` registered on instance A
+* There is no :term:`User` named ``"toto"`` who belongs to the ``"anonymous_network_A"`` group registered on instance B
+* There is a :term:`User` named ``"toto"`` who belongs to the ``"anonymous_network_A"`` group registered on instance C
+* Instance A is registered as a :term:`Network Node` on instances B and C
+* when ``"toto"`` gets a personal network token from instance A and uses it to log in on instance B they log in as the
+  the temporary ``"anonymous_network_A"`` user.
+* when ``"toto"`` gets a personal network token from instance A and uses it to log in on instance C they log in as the
+  ``"toto"`` user on instance C.
```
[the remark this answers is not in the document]
mishaschwartz: They are logged in at the `toto` user who belongs to the `toto_network_A` group

docs/authentication.rst
```diff
@@ -422,3 +422,84 @@ Furthermore, as described in the `procedure`_, :envvar:`MAGPIE_USER_REGISTRATION
...
+
+When another Magpie instance is registered as a :term:`Network Node`, a few additional entities are created:
+
+#. a group used to manage the permissions of all users who authenticate using the new :term:`Network Node`.
+   * this group's name will be the :envvar:`MAGPIE_NETWORK_NAME_PREFIX` followed by the :term:`Network Node` name
+#. a group used to manage the permissions of all users who authenticate using *any* other instance in the network
+   * this group's name will be the :envvar:`MAGPIE_NETWORK_GROUP_NAME`
+   * this group will only be created once, when the first :term:`Network Node` is registered
+#. an anonymous user that belongs to the two groups that were just created.
+   * this user name will be the :envvar:`MAGPIE_NETWORK_NAME_PREFIX` followed by the :term:`Network Node` name
```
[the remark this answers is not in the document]
mishaschwartz: If the `MAGPIE_NETWORK_NAME_PREFIX=pavics` then you shouldn't be able to create any new users whose name starts with `pavics`. Even so, the anonymous user for the `daccs` node would be `pavicsdaccs` (the prefix followed by the network name). 

Now, if you already had a user in the system named `pavicsdaccs` that could be a bit confusing but they should still be able to differentiate them because one would belong to the `pavicsdaccs` group and the original user would not.

If this is easier though, I'm happy to split these users out into different tables as you suggest (https://github.com/Ouranosinc/Magpie/pull/589#pullrequestreview-1627967830) if that will keep things cleaner.

magpie/adapter/magpieowssecurity.py
```diff
@@ -269,11 +269,18 @@ def update_request_cookies(self, request):
         """
         settings = get_settings(request)
         token_name = get_constant("MAGPIE_COOKIE_NAME", settings_container=settings)
+        network_mode = get_constant("MAGPIE_NETWORK_MODE", settings_container=settings,
+                                    settings_name="magpie.network_mode")
+        headers = dict(request.headers)
+        network_token_name = get_constant("MAGPIE_NETWORK_TOKEN_NAME", settings_container=settings)
+        if network_mode and "Authorization" not in headers and network_token_name in request.params:
+            headers["Authorization"] = "Bearer {}".format(request.params[network_token_name])
```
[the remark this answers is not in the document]
mishaschwartz: > Why not use the Authorization: Bearer {token} header directly?

I think the idea here was to provide an option for cases when a user wants to access a resource and doesn't have the ability to set the request headers. For example, if they're using a tool that just asks for a URL.

magpie/api/login/login.py
```diff
@@ -271,69 +279,127 @@ def login_success_external(request, external_user_name, external_id, email, prov
...
+                     .filter(models.NetworkToken.token == decoded_token["token"])
+                     .first())
+    if network_token:
+        return network_token.user
+
+
 @s.ProviderSigninAPI.get(schema=s.ProviderSignin_GET_RequestSchema, tags=[s.SessionTag],
                          response_schemas=s.ProviderSignin_GET_responses)
 @view_config(route_name=s.ProviderSigninAPI.name, permission=NO_PERMISSION_REQUIRED)
-def authomatic_login_view(request):
```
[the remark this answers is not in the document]
mishaschwartz: Yeah that's a great idea. I looked into it a bit but this is not my area of expertise at all. If you have some suggestions for where to start, what libraries to use, etc. then I can start looking into this.

An initial change with JWTs was much more feasible for me

magpie/api/login/login.py
mishaschwartz: The `user_from_token` function is also used in `validate_token_view`. These functions handle the exceptions differently depending on which `jwt.exceptions` class is raised. If I handle it in the `user_from_token` view and then return `None` then the calling functions are not able to handle different errors differently.

magpie/api/login/login.py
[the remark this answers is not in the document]
mishaschwartz: The main hurdle for me was how to set up authomatic providers that can be changed dynamically (ie. the same as creating a new `Network Node`)

magpie/api/management/network_node/network_node_views.py
mishaschwartz: Ok a few questions about this...

> For now, any admin could register a node, which could be updated/deleted/recreated at some other point by any other admin

Do you mean an admin for the same node or an admin for a different node? If you're talking about an admin for node A changing stuff of node B, I don't know how they would do that unless they had an account on both?

> If one node in the network gets hacked, that could allow a malicious user to generate tokens over the entire network.

Yes, I guess that if a node got hacked they could validate a token and return any user name they wanted. Not that that isn't bad and requires some thought, but I think that even with Oauth2, if the provider is compromised they can send any response they want anyway. Is this scenario different from that?

> This way, when a token is generated for a given node and sent to it, that node would have some ID to validate the token against and ensure it comes from a trusted location.

Right now, the JWT is validated to ensure that the token was issued by the right node. We could add another check to ensure that it was sent by one of the expected location(s). I think that's a great idea. Would that be something that using Oauth and authomatic would automatically get us?

magpie/alembic/versions/2023-08-25_2cfe144538e8_add_network_tables.py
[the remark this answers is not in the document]
fmigneault: In general, whenever `<user>.user_name` is accessed in the code, that `<user>` reference must be resolved beforehand based on what was passed as input from the API, UI, CLI or configuration startup.

This is an important one: 
https://github.com/Ouranosinc/Magpie/blob/1d602eedbe4047fd95c4d2684b5731f4d0600977/magpie/api/requests.py#L193

There are also some considerations to be taken for 
https://github.com/Ouranosinc/Magpie/blob/1d602eedbe4047fd95c4d2684b5731f4d0600977/magpie/models.py#L210
and 
https://github.com/Ouranosinc/Magpie/blob/1d602eedbe4047fd95c4d2684b5731f4d0600977/magpie/models.py#L921

which generates a somewhat related definition to a future user, but that is only really "linked" via `user_name` until the relevant registration/approval process is completed, since the actual `user_id` is not defined yet. 

This code never checks for potential multi-results returned from `user_name` matches because it always assumed the unique constraint. 

Similarly, all webhooks events for interactions with Cowbird work using `user_name`, since this is what the API paths use.

magpie/api/schemas.py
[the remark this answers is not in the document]
fmigneault: Until the WSGI app is returned from https://github.com/Ouranosinc/Magpie/blob/1d602eedbe4047fd95c4d2684b5731f4d0600977/magpie/app.py#L98
and actually started by Gunicorn, any intermediate `import ...` operation causes the imported module to be executed immediately before the "current app" that `get_constant` can detect is defined. 

Therefore, any call to `get_constant` at module level, such as in this case, is highly dependent on the order that modules are imported, which could be before or after the app is defined. Something that is working at one point could suddenly start breaking simply by changing imports, and this is hard to track. However, if you use `get_constant` within a function, the "current app" should exist because the function execution is performed during runtime, contrary to its definition.

docs/configuration.rst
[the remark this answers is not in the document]
fmigneault: Yes, that's fine. It's mostly a reminder for adjusting is as necessary when the feature is completed.

docs/authentication.rst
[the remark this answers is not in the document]
fmigneault: Ok. Will need to investigate more how conflicting user names are handled. Basically, we shouldn't assume that a name would necessarily be exactly the same on other Magpie instances. Preferably, alternative/remote Magpie instances should allow to "link" users of different names, similarly to how you can have 2 distinct user names on distinct social media platforms, but link accounts together. Retrieving the "resolved `user_name`" from another Magpie in the network could use similar approach to the `userinfo` of OAuth.

docs/authentication.rst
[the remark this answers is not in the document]
fmigneault: Yes, I think that would make it easier conceptually. The API endpoints however would still need to consider these conditions to allow user names or not. I think you mostly already did it with the regexes, but tests must be done to make sure it works everywhere (trying to purposely cause conflict, edge cases and such).

magpie/adapter/magpieowssecurity.py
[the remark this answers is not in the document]
fmigneault: Ok, I can understand the logic behind it. If really necessary for tools with limited capabilities, that can be permitted. 
However, consider that this is less secure than headers. Path/Query are too often dumped in logs or directly exposed. 
If you can use headers, it is preferable because it will obfuscate the tokens slightly more.

magpie/api/login/login.py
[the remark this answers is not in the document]
fmigneault: I think directly using `authomatic` by itself could work. It would take care of performing the OAuth requests flow with the relevant remote Magpie Network Node reference. 

The only thing that would be needed is to implement the endpoints to support OAuth 2.0, such as `/authorize`, `/userinfo`, etc. That is better explained in the spec https://datatracker.ietf.org/doc/html/rfc6749. 

I've not tried `authomatic` with JWT yet, but how to use them is well-defined in the standard (https://oauth.net/2/jwt-access-tokens/), so I would imagine it can support it natively. If not, it's maybe just a matter of adding the class that implements it, similar to the `esgfopenid` and `wso2` implementations in https://github.com/Ouranosinc/Magpie/blob/master/magpie/security.py and https://github.com/Ouranosinc/Magpie/tree/master/magpie/api/login

To make it dynamic with Magpie Network Nodes, I think expanding `authomatic_setup` with a set of configs passed as input would be sufficient. 

I've not played with the `authomatic` code for a while now. I would have to remind myself as well of all its features 😅

magpie/api/login/login.py
fmigneault: Ok.
The "how" to return the error needs to be given more though than just `None`. 
However, this is why I would prefer the try/except closer to the call that can raise it (I assume from `jwt.decode`). 
If `user_from_token` got refactored to handle exceptions differently, we might not notice that it break how it is used elsewhere. At least, I didn't notice until you pointed it out.

magpie/api/management/network_node/network_node_views.py
fmigneault: > Do you mean an admin for the same node or an admin for a different node? 

An admin of the node where the POST/PUT/DELETE requests are done. This is because of the default security check of Magpie that blocks requests from non-admin users. 

> Yes, I guess that if a node got hacked they could validate a token and return any user name they wanted. Not that that isn't bad and requires some thought, but I think that even with Oauth2, if the provider is compromised they can send any response they want anyway. Is this scenario different from that?

It wouldn't stop someone from sending requests from within the hacked node, but if an admin-user is compromised, it makes it slightly more complicated for someone to try to impersonate a fake Magpie node, because the other nodes that they could try to send token validations to would not have registered/handshake that fake node. It would also be easier to respond to some breach by revoking the hacked node client ID/secret. 

> Right now, the JWT is validated to ensure that the token was issued by the right node. We could add another check to ensure that it was sent by one of the expected location(s). I think that's a great idea. Would that be something that using Oauth and authomatic would automatically get us?

Yes. This is something like what I had in mind. However, the "location" doesn't need to be a URL, because that can be spoofed as well with requests. Doesn't hurt to have more validation layers though.
Yes, most of the features should be fairly well supported. It is used here for example https://github.com/Ouranosinc/Magpie/blob/master/magpie/security.py#L171-L190

docs/authentication.rst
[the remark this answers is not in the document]
mishaschwartz: Yeah I see what you mean. I'm feeling strongly from this review that we should be handling these remote users by storing them separately in another db table. That would resolve a lot of the issues that you raise here about conflicting names as well

docs/authentication.rst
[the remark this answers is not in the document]
mishaschwartz: Yeah I have some tests on the way but they're still in progress :)

magpie/adapter/magpieowssecurity.py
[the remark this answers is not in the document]
mishaschwartz: Yes, I completely agree. The short lifespan of these tokens is supposed to mitigate this risk a bit.

magpie/api/login/login.py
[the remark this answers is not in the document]
mishaschwartz: Ok this is super helpful. I'll start exploring this possibility this week and see how far I can get.

mishaschwartz:
@fmigneault 

I've been thinking about this:

> It seems like the custom external login code is essentially doing what authomatic would perform, but using non-standard OAuth methodology and endpoints.

I think that thinking about this from the perspective of OAuth is actually the wrong approach. Implementing a full OAuth provider is actually overkill for our purposes, but I think that we can adopt some of the OAuth ideas to improve the functionality and security of this new feature. 

The goal of this feature is simply to provide the user with an API token. This API token is analogous to OAuth's "access token" but:
-  we don't actually need the whole OAuth flow to acquire the token, the user can get the token directly from the node where they are registered. 
- we don't actually want the whole OAuth flow because that would require the user to authenticate with the authorization provider (the other node) every time they make a request to a new node (which is one thing we're trying to avoid here).
- the token should only be used to access resources protected by twitcher, this is _not_ intended to be used to log in to Magpie to access the Magpie UI for example.

We can still make the token exchange more secure by passing a client secret when one node passes the token to another.

Please let me know if you have any ideas for a different approach that I'm missing here

fmigneault:
@mishaschwartz 

> * we don't actually want the whole OAuth flow because that would require the user to authenticate with the authorization provider (the other node) every time they make a request to a new node (which is one thing we're trying to avoid here).

I'm not sure if I understand this part. When a UserNodeB wants to access a resource from NodeA using a token generated for (NodeB, UserNodeB) by NodeA, that token would need to be generated on behalf of that user after NodeA-NodeB handshake was accomplished.  Therefore, UserNodeB would need to be logged in NodeB in order to perform a "Request NodeA Token". That token request could be the "external provider login" procedure, which `authomatic` from NodeA would validate client id/secrets against the originating NodeB by calling and counter-validating the `/userinfo` the UserNodeB claimed to be. The only distinction of this procedure from the current code to support Network Mode would be that an access token would be returned to NodeB instead of the `Cookie` currently used to complete log in NodeA.

Did you have another procedure in mind? I don't see how else the nodes can validate that the user requesting an access token is legitimate if NodeA cannot re-validate UserNodeB on NodeB with an appropriate login.

> * the token should only be used to access resources protected by twitcher, this is _not_ intended to be used to log in to Magpie to access the Magpie UI for example.

Technically speaking, once Magpie can authenticate the user, it gives the same access via Twitcher. There are no distinction between these two. When logged in an "admin", you obtain access to all Twitcher services and Magpie UI pages. When non-admin, you get restricted Twitcher access based on Magpie permissions and limited access to Magpie UI to `/magpie/ui/users/current`. Therefore, if a token would authorize a user to access a resource via Twitcher, that same token could be used to log in Magpie.

mishaschwartz:
> Did you have another procedure in mind? 

Yeah... 

- UserNodeB would like to access a resource on NodeA
- UserNodeB logs in and acquires a token from NodeB
- UserNodeB makes a request to NodeA and includes the token
- NodeB sends the token to NodeA for validation
- if the token is valid, NodeA responds with information about UserNodeB (their user_name)
- NodeB validates that the response is from a trusted source
- NodeB makes an authorization decision based on the response (whether to give UserNodeB access to the requested resource)


Just to clarify a few things:

> When a UserNodeB wants to access a resource from NodeA using a token generated for (NodeB, UserNodeB) by NodeA

The token is generated by NodeB for UserNodeB, not by NodeA

> UserNodeB would need to be logged in NodeB in order to perform a "Request NodeA Token"

The token is not for NodeA, the token is generated and validated exclusively by NodeB. The same token can be included in a request to any node (NodeA, NodeB, NodeC, NodeD) and it will still work.

fmigneault:
> > When a UserNodeB wants to access a resource from NodeA using a token generated for (NodeB, UserNodeB) by NodeA
> 
> The token is generated by NodeB for UserNodeB, not by NodeA

This is the problem. 
The resources are protected by NodeA, so it is NodeA that must generate the token. 
It cannot be the other way around, otherwise the token cannot be secured and signed by NodeA's secret, which NodeB cannot ever know. If NodeB knew that secret, that would make the full network at risk because it would assume every node knows every other node's secrets bidirectionally.

You are missing a key step here: 
- UserNodeB logs in and acquires a token from NodeB
-  *** NodeB obtains an access token from NodeA for validated UserNodeB ***
- UserNodeB makes a request to NodeA and includes the token


> The token is not for NodeA

Seems to contradict the first point: "UserNodeB would like to access a resource on NodeA".

No matter how the token/cookie is generated, when Twitcher validates with Magpie for AuthN/AuthZ of said token, it will be for accessing resources on its own instance. Twitcher/Magpie of NodeA do not care about how NodeB resources are accessed and by whom. It is the problem of this instance's Twitcher/Magpie.

> The same token can be included in a request to any node (NodeA, NodeB, NodeC, NodeD) and it will still work.

That is just not possible, by concept, for security. Each node should have respective and distinct secret signing of their own tokens. 
If not, that bypasses the purpose of encrypting the tokens. It's just like passing around your private SSH key to anyone that want to access the protected instance, and then "trust" them not to share it around for other (malicious) purpose.

mishaschwartz:
Ok, just to be clear, I understand what you're saying. You're describing a solution using Oauth, I'm describing one that doesn't use Oauth but relies on limited resource access and short token lifespans to mitigate unauthorized access.  

The main advantage of my solution is that it requires fewer steps for users and node administrators to set up a system that allows them to access resources on multiple nodes.

BUT ... after reviewing your proposed solution again, I'm happy to go with your proposal and we can provide the user with client-side tools to simplify their workflow instead. I don't know what that will look like yet but I'm sure we can figure something out.

I think the most important thing is to get this working in the short time that CRIM is still available for the project.

fmigneault:
> The main advantage of my solution is that it requires fewer steps for users and node administrators to set up a system that allows them to access resources on multiple nodes.

So it does for someone trying to bypass security! Hence why I see it on the flip side as a disadvantage. 

The main issue I can see with the proposed solution was that it uses a less secure approach by design. If it was a security method on its own, it wouldn't be as problematic, but since it inevitably ties back into the Twitcher authorization at some point, that new method would create the weakest link security-wise affecting the overall solution.

I'm not too strict about using Oauth or an alternative method, as long as the core principles around it align.

> provide the user with client-side tools to simplify their workflow instead

I believe this could be made relatively transparent for the user. It is rather the Magpie nodes themselves that would need to perform a few more request validation steps to sync tokens between each other behind the scene. Once Cookie or Authorization headers are set, the user shouldn't see the difference where the resources are being accessed from.

mishaschwartz:
@fmigneault 

I have two new proposals for the access flow. 

Both options are similar to OAuth's client credentials flow with the added protection that both Nodes validate the requests/responses of each other using asymmetric public/private keys (can be implemented by passing RSA256 signed JWTs)

Option 1: UserA acquires a token to access a specific resource on a remote NodeB
Option 2: UserA acquires a token to access the remote NodeB as a specific user on NodeB (anonymous or otherwise)

For both proposals, we assume the following has already been set up:

- UserA has an account on NodeA
- UserA may or may not have an account on NodeB
- All nodes have a public/private key pair
- NodeA has NodeB's public key
- NodeB has NodeA's public key

Access flow for Option 1:

- UserA logs in to NodeA
- UserA sends a request to NodeA to get an access token for a resource on NodeB
    - ex: https://nodea.com/network/nodeb?resource=/thredds/datasets/example.ncml
- NodeA sends a request for an access token to NodeB:
    - ex: https://nodeb.com/token?jwt=jwtstringgoeshere-signedwithNodeAprivatekey 
          - content of JWT (signed with NodeA's private key):
            {"client": "nodeA", "user": "userA", "resource": "/thredds/datasets/example.ncml"}
- NodeB receives the request and verifies the JWT with NodeA's public key
- NodeB check's if userA is authorized to access the resource in question
    - if yes: 
        - NodeB creates a new token in its database with the following values:
            - random string (UUID)
            - resource_id
        - NodeB responds with JWT (signed with NodeB's private key)
            - content: {"uuid": "random-string-from-database", "exp": "expiry-date"}
    - if no:
        - NodeB responds with 403
- NodeA validates the JWT with NodeB's public key and responds to the user with the JWT (if valid)
- UserA sends a request to NodeB to access the resource and includes the JWT:
    - ex: https://nodeb.com/thredds/datasets/example.ncml?token=jwt-accesss-token
    - ex2: https://nodeb.com/thredds/datasets/example.ncml (and include the JWT in the header)
- NodeB validates the JWT using its own public key, checks that the UUID exists in the database and matches the resource in the database
    - if yes:
        - NodeB continues to resource access
    - if no
        - NodeB responds with 403

Access flow for Option 2:

- UserA logs in to NodeA
- UserA sends a request to NodeA to get an access token for NodeB
    - ex: https://nodea.com/network/nodeb
- NodeA sends a request for an access token to NodeB:
    - ex: https://nodeb.com/token?jwt=jwtstringgoeshere-signedwithNodeAprivatekey 
          - content of JWT (signed with NodeA's private key):
            {"client": "nodeA", "user": "userA"}
- NodeB receives the request and verifies the JWT with NodeA's public key
- NodeB matches UserA with a user that already exists on NodeB (can be one a specific user or the "NodeA anonymous user")
    - NodeB creates a new token in its database with the following values:
        - random string (UUID)
        - user_id
- NodeB responds with JWT (signed with NodeB's private key)
    - content: {"uuid": "random-string-from-database", "exp": "expiry-date"}
- NodeA validates the JWT with NodeB's public key and responds to the user with the JWT (if valid)
- UserA sends a request to NodeB to access a resource and includes the JWT:
    - ex: https://nodeb.com/thredds/datasets/example.ncml?token=jwt-accesss-token
    - ex2: https://nodeb.com/thredds/datasets/example.ncml (and include the JWT in the header)
- NodeB validates the JWT using its own public key, checks that the UUID exists in the database and checks that the user has permission to access the resource
    - if yes:
        - NodeB continues to resource access
    - if no
        - NodeB responds with 403


Let me know if you have any questions or an opinion about which method you'd prefer.

Some security implications to consider:

- In both options an attacker who steals a token would need to already know which Node issued which token in order to gain access.
- In Option 1, an attacker would *also* need to know the precise resource that the token was issued for.
- In Option 1, a user would need to get a unique token for every resource they want to access
- In Option 2, a user only needs one token to access multiple resources
- In Option 1, a user could acquire a token, then an admin could revoke permission to that resource but the token would still provide access (until it expires or is revoked)

fmigneault:
## Option 1: 

> NodeB check's if userA is authorized to access the resource in question

How does NoteB perfoms that check? 
Since there are no "UserB" with that option, there is no permission defined for `/thredds/datasets/example.ncml`.
There is also no possible resolution of group permissions (directly on the resource or by resource hierarchy resolution logic), since no "UserB" can be defined as a member of a given group. The "effective permission" is resolved using users and their group memberships currently, but not groups on their own. 

## Option 2: 

> NodeB matches UserA with a user that already exists on NodeB (can be one a specific user or the "NodeA anonymous user") 

That could work, but we need to figure out how we will define the user to avoid potential name conflicts.

One thinng to consider is that, if UserA is not admin on NodeB (shouldn't be, otherwise they wouldn't need the token), they cannot list users on NodeB. Therefore, they have limited capabilities to "associate" their UserA to some "NodeBUserA". Trusted NodeB by NodeA could do this association on behalf of UserA, but again, we must consider how to handle name conflicts. 

-----

In both options, the rest of the procedure seem adequate.

I tend to lean toward __*Option 2*__ more, because that involves less (no) modifications in the "effective permission" logic. There's also less API/UI to develop to support "admin revoking a resource token", since it can use the current serivce/permission interface. After resolving the JWT token into some auth/cookie, Magpie/Twitcher would see the resolved "NodeBUserA" as any other local NodeB user (and applicable groups if need be).

> In Option 2, a user only needs one token to access multiple resources

This is also a compelling argument for that option. It allows the user to directly interact with all its resources without having to manage individual tokens per endpoint, which would be very complicated if they try to do so themselves (eg: when accessing data in a notebook). 

## Extra Considerations

Keys of remote nodes would have to be stored in DB using an extra hashing pass using the secret of the local instance.
This way tokens will not be stored in "plain text".

fmigneault:
Ok. Sounds good for admin creating the associations for now, and having a separate user table. 
The user self-association is also a good idea, but indeed better to keep it for a future PR. 

Now, considering that separate user table, how would the admin make that association?
There is a need for API endpoint(s) (at least), and potentially a new UI page, to allow them to add/list/remove remote users and the associations between local/remote users.

mishaschwartz:
> There is a need for API endpoint(s) (at least), and potentially a new UI page, to allow them to add/list/remove remote users and the associations between local/remote users.

Yes for sure. The plan is to start with the API endpoints and then add the UI additions as a second step. 

The plan for the UI is to add a new field to the form on `/ui/users/{username}/default` to add/remove a remote username and node name (username is a text field, node name is a drop-down list).

Then we can add a new page to list all users with associated remote user names (admin view only). We could also display this information on `/ui/users` if you'd prefer to not create a new page but that might get messy since a single user can have multiple associations to multiple remote users on other nodes in the network.

mishaschwartz:
@fmigneault 

All api and ui changes have now been made and all existing test and code check are now passing. 

I will start writing tests for the new functionality but the code and doc changes are ready for another look.

magpie/ui/network/views.py
```diff
@@ -0,0 +1,45 @@
...
+        # Extend this to other response types later if needed
+        if response_type != "id_token":
+            raise HTTPBadRequest("Invalid response type")
+        if token is None:
+            raise HTTPBadRequest("Missing token")
+        try:
+            node_name = jwt.decode(token, options={"verify_signature": False}).get("iss")
+        except jwt.exceptions.DecodeError:
+            raise HTTPBadRequest("Token is improperly formatted")
+        node = self.request.db.query(NetworkNode).filter(NetworkNode.name == node_name).first()
```
fmigneault, line 30: Move DB operations to API.

magpie/ui/network/views.py
```diff
@@ -0,0 +1,45 @@
...
+        response_type = self.request.GET.get("response_type")
+        redirect_uri = self.request.GET.get("redirect_uri")
+
+        # Extend this to other response types later if needed
+        if response_type != "id_token":
+            raise HTTPBadRequest("Invalid response type")
+        if token is None:
+            raise HTTPBadRequest("Missing token")
+        try:
+            node_name = jwt.decode(token, options={"verify_signature": False}).get("iss")
```
fmigneault, line 27: I think the JWT checks and encoding strategies are too specific to be done on the UI side. 
They should all be done within the API to avoid duplicating this logic, and have a way to do the same directly with API requests.

We should limit quick pre-checks only to obvious issues such as the bad response type or empty token above to avoid refreshing the whole UI page only for an error, which will cause more requests/processing than simply returning the error here.

magpie/models.py
```diff
@@ -1059,18 +1095,31 @@ class NetworkNode(BaseModel, Base):
...
+        return self.anonymous_user_name_formatter(self.name)
 
-    @property
-    def anonymous_user(self):
-        # type: () -> Optional[User]
-        return self.users.filter(User.user_name == self.anonymous_user_name).first()
+    @staticmethod
+    def anonymous_user_name_formatter(name):
+        # type: (Str) -> Str
+        return "{}{}".format(get_constant("MAGPIE_NETWORK_NAME_PREFIX"), name)
```
fmigneault, line 1110: I just got an idea regarding this to avoid possible conflicts or misinterpretation of user names when reading responses.
How about using explicitly a disallowed character to separate the node prefix from the user name?
Something like `{node-name}:{user-name}`. 
This way, we would never have to worry about misinterpretation or conflicts of network user names.

magpie/models.py
```diff
@@ -1042,13 +1040,51 @@ class NetworkToken(BaseModel, Base):
...
+        self.created = datetime.datetime.utcnow()
+        return unhashed_token
+
+    def expired(self):
+        # type: () -> bool
+        expiry = int(get_constant("MAGPIE_NETWORK_DEFAULT_TOKEN_EXPIRY"))
+        return (datetime.datetime.utcnow() - self.created) > expiry
+
+    @classmethod
+    def get_expired(cls, db_session):
```
fmigneault, line 1070: Unless I didn't see properly, the only use is in the CLI to call `delete()`. Would be better to simply have `delete_expired` instead to avoid possible mishandling by the CLI.

magpie/models.py
```diff
@@ -1034,6 +999,39 @@ def json(self):
         return {"token": str(self.token), "operation": str(self.operation.value)}
 
 
+class NetworkRemoteUser(BaseModel, Base):
+    """
+    Model that defines a relationship between a User and a User that is authenticated on a different NetworkNode.
```
fmigneault, line 1004: Use RST `:class:` marker to relevant references.

magpie/cli/purge_expired_network_tokens.py
```diff
@@ -0,0 +1,70 @@
+#!/usr/bin/env python3
+
+"""
+Removes expired network tokens from the database.
```
fmigneault, line 4: Seems like a useful operation that should be directly in the API (eg: `DELETE /network/tokens` or similar). 
The CLI could simply call that operation and provide the errors/success based on what the API responds.

magpie/api/schemas.py
```diff
@@ -3638,6 +3526,367 @@ class SwaggerAPI_GET_OkResponseSchema(colander.MappingSchema):
...
+        description="URL that provides the Oauth token endpoint for another Magpie node (instance) in the network.",
+        example="https://nodea.example.com/network/token",
+        validator=colander.url,
+        missing=colander.drop
+    )
+    redirect_uris = colander.SchemaNode(
+        colander.String(),
+        description="Space delimited list of valid redirect URIs for another Magpie node (instance) in the "
+                    "network.",
+        example="https://node.example.com/network/nodes/link https://node.example.com/some/other/uri",
```
fmigneault, line 3577: Wrong values depending on previous comments.

magpie/api/schemas.py
```diff
@@ -3638,6 +3526,367 @@ class SwaggerAPI_GET_OkResponseSchema(colander.MappingSchema):
...
+    token_url = colander.SchemaNode(
+        colander.String(),
+        description="URL that provides the Oauth token endpoint for another Magpie node (instance) in the network.",
+        example="https://nodea.example.com/network/token",
+        validator=colander.url,
+    )
+    redirect_uris = colander.SchemaNode(
+        colander.String(),
+        description="Space delimited list of valid redirect URIs for another Magpie node (instance) in the network",
+        example="https://node.example.com/network/nodes/link https://node.example.com/some/other/uri",
```
fmigneault, line 3618: Wrong values depending on previous comments.

magpie/api/schemas.py
```diff
@@ -3638,6 +3526,367 @@ class SwaggerAPI_GET_OkResponseSchema(colander.MappingSchema):
...
+class NetworkNodeLink_GET_OkResponseSchema(BaseResponseSchemaAPI):
+    description = "User successfully linked with network node."
+    body = BaseResponseBodySchema(code=HTTPOk.code, description=description)
+
+
+class NetworkRemoteUser_PATCH_RequestBodySchema(colander.MappingSchema):
+    remote_user_name = colander.SchemaNode(
+        colander.String(),
+        description="Name of the associated user a remote Magpie node (instance) in the network.",
+        example="userAremote",
```
fmigneault, line 3710: Maybe wrong example value depending on prefix name idea.

magpie/api/schemas.py
```diff
@@ -3638,6 +3526,367 @@ class SwaggerAPI_GET_OkResponseSchema(colander.MappingSchema):
...
+class NetworkRemoteUsers_GET_NotFoundResponseSchema(BaseResponseSchemaAPI):
+    description = "Network User could not be found."
+    body = ErrorResponseBodySchema(code=HTTPNotFound.code, description=description)
+
+
+class NetworkRemoteUser_BodySchema(colander.MappingSchema):
+    remote_user_name = colander.SchemaNode(
+        colander.String(),
+        description="Name of the associated user a remote Magpie node (instance) in the network.",
+        example="userAremote"
```
fmigneault, line 3750: Maybe wrong example value depending on prefix name idea.

magpie/api/management/user/user_utils.py
```diff
@@ -853,8 +851,8 @@ def get_user_service_resources_permissions_dict(user, service, request,
 
 
 def check_user_info(user_name=None, email=None, password=None, group_name=None,  # required unless disabled explicitly
-                    check_name=True, check_email=True, check_password=True, check_group=True):
-    # type: (Str, Str, Str, Str, bool, bool, bool, bool) -> None
+                    check_name=True, check_email=True, check_password=True, check_group=True, check_anonymous=True):
```
fmigneault, line 854: Not against regrouping the check under this call, but I got a vague memory that the reserved anonymous checks were done where needed in the code. Potentially duplicate code could be removed in favor of passing this argument?

docs/configuration.rst
```diff
@@ -1059,11 +1069,40 @@ to authenticate users for each other. All variables defined in this section are
...
+
+    [:class:`str`]
+    (Default: ``None``)
+
+    .. versionadded:: 3.38
+
+    Password used to encrypt the PEM files in :envvar:`MAGPIE_NETWORK_PEM_FILES`.
+
+    If multiple files require passwords, they can be listed with a ``:`` character separator (Note that this means that
+    passwords cannot contain a ``:``. An empty string will be treated the same as no password.
```
fmigneault, line 1101: Could require that passwords are `base64` encoded to avoid this problem.

fmigneault (changes requested):
First pass of review. Didn't complete checking everything.
Will need more reviews after tests are added and adjustments applied. 

I'm considering using a `4.0` version for this. The feature is sufficiently important to justify that upgrading could need additional considerations by node admins.

magpie/api/management/network/network_views.py
```diff
@@ -0,0 +1,48 @@
...
+        request.db.delete(network_remote_user.network_token)
+        if (network_remote_user.user.id == node.anonymous_user(request.db).id and
+                sqlalchemy.inspect(network_remote_user).persisted):
+            request.db.delete(network_remote_user)  # clean up unused record in the database
+        return ax.valid_http(http_success=HTTPOk, detail=s.NetworkToken_DELETE_OkResponseSchema.description)
+    ax.raise_http(http_error=HTTPNotFound, detail=s.NetworkNodeToken_DELETE_NotFoundResponseSchema.description)
+
+
+@s.NetworkJSONWebKeySetAPI.get(tags=[s.NetworkTag], response_schemas=s.NetworkJSONWebKeySet_GET_responses)
+@view_config(route_name=s.NetworkJSONWebKeySetAPI.name, request_method="GET", permission=NO_PERMISSION_REQUIRED)
```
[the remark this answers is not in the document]
mishaschwartz: This route provides the public json web key for the current node. This is the public half of a public/private key pair. It is used by other nodes to verify a json web token that was signed by this nodes private key.

To mimic this node, they would need access to the nodes private key (which is not and never should be accessible).

In a previous version of this PR, we were using a symmetric signing algorithm (HS256) but now we're using the asymmetric RS256 algorithm instead.

docs/authentication.rst
```diff
@@ -428,44 +428,76 @@ their services.
...
+    * This is usually ``https://{hostname}/network/jwks`` where ``{hostname}`` is the hostname of the other instance
+* ``authorization_url``
+    * URL that provides the instance's Oauth authorize endpoint.
+    * This is usually ``https://{hostname}/ui/network/authorize`` where ``{hostname}`` is the hostname of the other
+      instance
+* ``token_url``
+    * URL that provides the instances Oauth token endpoint.
+    * This is usually ``https://{hostname}/network/token`` where ``{hostname}`` is the hostname of the other instance
+* ``redirect_uris``
+    * Space delimited list of valid redirect URIs for the instance. These are used by the instance's Oauth authorize
```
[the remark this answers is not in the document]
mishaschwartz: why not? I'll change it :)

docs/authentication.rst
```diff
@@ -428,44 +428,76 @@ their services.
...
+      endpoint to safely redirect the user back once they have authorized `Magpie` to link their accounts on two
+      different instances.
+    * This is usually ``https://{hostname}/network/nodes/link`` where ``{hostname}`` is the hostname of the other
```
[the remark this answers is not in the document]
mishaschwartz: Good catch. I'll update it to use `/network/link`

magpie/constants.py
```diff
@@ -197,11 +202,32 @@ def protected_user_name_regex(include_admin=True,
...
-    if include_network and get_constant("MAGPIE_NETWORK_MODE", settings_container=settings_container):
+    if include_network and network_enabled(settings_container=settings_container):
         patterns.append(
             "{}.*".format(get_constant("MAGPIE_NETWORK_NAME_PREFIX", settings_container=settings_container))
         )
-    return "^{}$".format("|".join(patterns))
+    return "^({})$".format("|".join(patterns))
+
+
+def protected_user_email_regex(include_admin=True,
```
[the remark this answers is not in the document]
mishaschwartz: What caching function do you recommend? I like `functools.cache` but that's not available in versions < 3.9

Or do you mean manually cache with global variables?

magpie/ui/network/__init__.py
```diff
@@ -0,0 +1,11 @@
+from magpie.utils import get_logger
+
+LOGGER = get_logger(__name__)
+
+
+def includeme(config):
+    from magpie.ui.network.views import NetworkViews
+    LOGGER.info("Adding UI network...")
+    path = "/ui/network/authorize"
+    config.add_route(NetworkViews.authorize.__name__, path)
```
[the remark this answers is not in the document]
mishaschwartz: I'm confused...

This is a UI endpoint, not API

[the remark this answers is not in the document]
fmigneault: You're right. I got mixed by the `add_route` using `NetworkViews` and incorrectly assumed it was a CorniceSwagger service. 
Just in case, can you adjust `NetworkViews.authorize.__name__` by `fully_qualified_name(NetworkViews.authorize)` so that it does not conflict with other "authorize" if more are added later?

magpie/constants.py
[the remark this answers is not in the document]
fmigneault: You can use `functiools.lru_cache`. `functools.cache` is a shortcut to `functiools.lru_cache(None)`.

magpie/ui/network/templates/authorize.mako
```diff
@@ -0,0 +1,32 @@
...
+
+<form id="authorize_node_link"
+      action="${authorize_uri}"
+      class="new-item-form"
+      method="get">
+    <table class="authorization-table">
+        <tr>
+            <td>
+                Magpie is requesting permission to link your account with the ${requesting_user_name} user's account
+                on the Magpie instance named "${node_name}".
```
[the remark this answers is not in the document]
mishaschwartz: It's a good idea. But we don't actually have the canonical URL for the nodes stored anywhere.

It's not needed since we have the various endpoints stored instead. Since this is just for display purposes, I could just use the referrer URL, which is the URL the user should expect to see anyway (since that was where they would have initiated the link request from).

magpie/ui/network/views.py
mishaschwartz: Alright, I'll try to reorganize a bit and see how it looks

magpie/models.py
mishaschwartz: I don't think this is a good idea. 

The user name validation checks still run when we create or update this user so we'd have to explicitly skip those when creating these users. 

That by itself might be ok, but this would also mean that downstream applications (like jupyterhub for example) wouldn't be able to rely on the user names to always be properly formatted and we might run into another problem like the one in https://github.com/bird-house/birdhouse-deploy/pull/396

magpie/cli/purge_expired_network_tokens.py
mishaschwartz: I agree that it's useful to have. 

I'll add it in the API but I'd like to keep the option to perform the deletion on the DB directly because that way we can simplify the crontab entry by not requiring a username and password.

magpie/api/management/user/user_utils.py
mishaschwartz: I've replaced most of the previous anonymous checks because they no longer apply.

Previously, they were just checking whether it matched the reserved anonymous username. Now it checks against the regular expression defined by the `protected_user_name_regex` function

docs/configuration.rst
mishaschwartz: That seems like a lot of extra work for the admin user.

What if I just format it as a json list (like you suggested for the redirect_uris above)

docs/authentication.rst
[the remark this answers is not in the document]
mishaschwartz: Remote users are now stored in a different table: `network_remote_users`

magpie/api/login/login.py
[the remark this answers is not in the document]
mishaschwartz: I've moved the code to a separate function

magpie/cli/purge_expired_network_tokens.py
mishaschwartz: Added an option to do this with an API call

fmigneault:
> please see updated documentation

I think you read my mind 😆 
I was just thinking this is something that should be represented in a schema to facilitate interpretation instead of reviewing the code each time. 

I'll try to dedicate some time next week to look at it in detail.

docs/_static/client-request-flow.drawio
fmigneault: Point (1) says it requests Node A, but the link shows a request toward Node B.
It is not really clear from that schema what is the actual interaction with Node A.

The resource (`/some/proteceted/resource`) assumes something else (e.g.: Twitchier) is involved. While that makes senses, it is not really obvious from a Magpie-only perspective what this represents, since there is no such endpoint in Magpie's API.

The text mentioning it "*assumes that the user is not logged in to Node B*" might be misleading or hard to interpret, since an `Authorization` header is provided, and that could be considered a "login token", e.g.: auto-login from `Authorization` is handled in: https://github.com/Ouranosinc/Magpie/blob/4faf23d485983fd23c7f4f4e74d61d48bc5c3a8d/magpie/adapter/magpieowssecurity.py#L261

Does "logged in" in this case assumes no "cookie" or something else? It must be clarified.

It is also not really clear how the resolution between Node A/B happens. While there is a `provider: magpie_network`, it does not identify which potential Node (C, D, etc.?) it should like against. The token should not be sent arbitrarily to all potential Node candidates to reduce chances of leaking it.

.github/workflows/tests.yml
```diff
@@ -62,11 +62,18 @@ jobs:
...
             python-version: "3.11"
             allow-failure: true
             test-case: start test-remote
+          # remote network tests using a remote server with network mode enabled
+          - os: ubuntu-latest
+            python-version: "3.12"
+            allow-failure: true
+            test-case: start test  # the tests that are run are limited by the PYTEST_ADDOPTS in test-option below
+            test-option: >- 
+              PYTEST_ADDOPTS='-m "remote and network"' MAGPIE_NETWORK_CREATE_MISSING_PEM_FILE=True MAGPIE_NETWORK_ENABLED=on MAGPIE_NETWORK_INSTANCE_NAME=example
```
fmigneault: While this works functionality-wise, the CI displays it not that nicely. It might be worth having a dedicated `test-remote-network` target that does these steps to facilitate running it locally and understand the CI matrix operations more easily.

Image

Image

.github/workflows/tests.yml
```diff
@@ -105,6 +112,9 @@ jobs:
         run: |
           hash -r
           env | sort
+      - name: Create private key for network testing
+        if: ${{ matrix.test-case == 'test-local' }}
+        run: ${{ matrix.test-option }} make create-private-key
```
fmigneault: Is this for non-remote network unit tests? Remote network tests does not need it as well?
Are non-remote network tests expected to run along other unit tests? What if network mode is disabled for that instance? They will automatically fail?

magpie/constants.py
```diff
@@ -146,6 +154,11 @@ def _get_default_log_level():
 MAGPIE_CONTEXT_PERMISSION = "MAGPIE_CONTEXT_USER"  # path user must be itself, MAGPIE_LOGGED_USER or unauthenticated
 MAGPIE_LOGGED_USER = "current"
 MAGPIE_DEFAULT_PROVIDER = "ziggurat"
+MAGPIE_NETWORK_TOKEN_NAME = "magpie_token"  # nosec: B105
+MAGPIE_NETWORK_PROVIDER = "magpie_network"
+MAGPIE_NETWORK_NAME_PREFIX = "anonymous_network_"
```
fmigneault: Not blocker, but since `anonymous` is simply the default not-logged-in name, and could be set to a different value, this could lose its meaning for nodes changing it. 

Are there assumptions that each node will use `anonymous` when referring to this concept?

CHANGES.rst
```diff
@@ -40,7 +40,11 @@ Features / Changes
 ------------------------------------------------------------------------------------
 
 Features / Changes
-~~~~~~~~~~~~~~~~~~~~~
+~~~~~~~~~~~~~~~~~~
+
+* Introduce "Network Mode" which allows other Magpie instances to act as external authentication providers using access
+  tokens. This allows users registered across multiple Magpie instances in a network to more easily gain access to the
+  resources within the network, without requiring the duplication of user credentials across the network.
```
fmigneault: needs to be moved up

docs/authentication.rst
```diff
@@ -422,3 +422,176 @@ Furthermore, as described in the `procedure`_, :envvar:`MAGPIE_USER_REGISTRATION
...
 server managers that use `Magpie` to decide if they desire more control over which individuals can join and access
 their services.
+
+.. _Network Mode:
+
+Network Mode
+------------
+
+If the :envvar:`MAGPIE_NETWORK_ENABLED` is enabled, `Magpie` instances can be linked in a network which allows them to
+associate user accounts across the network and provide limited resource access to users who have accounts on other
```
fmigneault: Use ``` :term:`User` ```, ``` :term:`Resource` ``` at least for the first references.

docs/glossary.rst
```diff
@@ -138,6 +138,23 @@ Glossary
         :py:data:`magpie.constants.MAGPIE_ANONYMOUS_USER`. Otherwise, it is whoever the
         :term:`Authentication` mechanism identifies with token extracted from request :term:`Cookies`.
 
+    Network Node
+        A reference to an instance of the Magpie software within a network of Magpie instances. Each Magpie instance
```
fmigneault: ``` `Magpie` ```

docs/glossary.rst
```diff
@@ -138,6 +138,23 @@ Glossary
...
+        within the network is registered in the database as a row in the ``network_nodes`` table. Each node is
+        represented by a name that is unique across all nodes in the network, and a url that is used to send http
```
fmigneault: `URL` and `HTTP`

docs/configuration.rst
```diff
@@ -986,6 +986,134 @@ remain available as described at the start of the :ref:`Configuration` section.
     Name of the :term:`Provider` used for login. This represents the identifier that is set to define how to
     differentiate between a local sign-in procedure and a dispatched one some known :ref:`authn_providers`.
 
+Network Mode Settings
+~~~~~~~~~~~~~~~~~~~~~
+
+The following configuration parameters are related to Magpie's "Network Mode" which allows networked instances of Magpie
```
fmigneault: ``` :ref:`network_mode` ``` here and below

``` `Magpie` ``` here and below. Also consider replacing by more specific terms when possible, such as reusing the `Network Node` term instead of `Magpie instance`.

File: docs/configuration.rst [hunk 39]
```diff
@@ -986,6 +986,134 @@ remain available as described at the start of the :ref:`Configuration` section.
...
+    Enable "Network Mode" which enables all functionality to authenticate users using other Magpie instances as
+    external authentication providers.
+
+.. envvar:: MAGPIE_NETWORK_INSTANCE_NAME
+
+    [:class:`str`]
+
+    .. versionadded:: 3.38
+
+    The name of this Magpie instance in the network. This variable is used to determine if an authentication token was
```
fmigneault: term `Authentication`

docs/configuration.rst
```diff
@@ -986,6 +986,134 @@ remain available as described at the start of the :ref:`Configuration` section.
...
+
+.. envvar:: MAGPIE_NETWORK_PROVIDER
+
+    [|constant|_]
+    (Value: ``"magpie_network"``)
+
+    .. versionadded:: 3.38
+
+    The name of the external provider that authenticates users using other Magpie instances as external authentication
+    providers.
```
fmigneault: use existing terms about external providers and auth

docs/configuration.rst
```diff
@@ -986,6 +986,134 @@ remain available as described at the start of the :ref:`Configuration` section.
...
+
+.. envvar:: MAGPIE_NETWORK_NAME_PREFIX
+
+    [|constant|_]
+    (Value: ``"anonymous_network_"``)
+
+    .. versionadded:: 3.38
+
+    A prefix added to the anonymous network user and network group names. These names are constructed by prepending the
```
fmigneault: Reuse term `Network User`.

Should there be a `Network Group` term as well? It seems only related to `MAGPIE_NETWORK_GROUP_NAME`. Is that correct? What is its implication/usage, and could it be needed to have different onces per node?

docs/configuration.rst
```diff
@@ -986,6 +986,134 @@ remain available as described at the start of the :ref:`Configuration` section.
...
+
+    .. versionadded:: 3.38
+
+    The name of the external provider that authenticates users using other Magpie instances as external authentication
+    providers.
+
+.. envvar:: MAGPIE_NETWORK_NAME_PREFIX
+
+    [|constant|_]
+    (Value: ``"anonymous_network_"``)
```
fmigneault: I'm not sure to understand why `anonymous` is used here specifically and not referenced in the variable name. Is it only for this special user/group? If so, it would be better to have explicit `MAGPIE_NETWORK_ANONYMOUS_NAME_PREFIX`. It would also be better to use `magpie_network_anonymous_` to align with other `magpie_network...` names.

magpie/cli/create_private_key.py
fmigneault: Missing documentation in `utilities.rst`.

magpie/cli/purge_expired_network_tokens.py
fmigneault: Missing documentation in `utilities.rst`.

magpie/api/management/user/user_utils.py
```diff
@@ -858,8 +854,8 @@ def get_user_service_resources_permissions_dict(user, service, request,
...
```
fmigneault: Could the parameter be adjusted to `check_not_anonymous` to improve readability?
Looking at the code invoking this function, `check_anonymous=True` gives the impression that it should "*check the value matches anonymous*", while the opposite check is actually performed.

magpie/api/requests.py
```diff
@@ -379,3 +381,27 @@ def get_query_param(request, case_insensitive_key, default=None):
             if param.lower() == key.lower():
                 return request.params.get(param)
     return default
+
+
+def check_network_mode_enabled(view_func):
+    # type: (Callable) -> Callable
```
fmigneault: Can you transfer the definitions from https://github.com/crim-ca/weaver/blob/cba028a11d1bee42d5b7c81211ad6ec8b1a717bf/weaver/typedefs.py#L377-L386

and apply them along https://github.com/Ouranosinc/Magpie/blob/b589ab11751e58e97d55c0affb7661b1fb9d5b00/magpie/typedefs.py#L51-L55

to reuse `AnyViewCallable` here

magpie/cli/create_private_key.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+def make_parser():
+    # type: () -> argparse.ArgumentParser
+    parser = argparse.ArgumentParser(description="Create a private key used to generate a JSON Web Key.")
+    parser.add_argument("--config", "--ini", metavar="CONFIG", dest="ini_config",
+                        default=get_constant("MAGPIE_INI_FILE_PATH"),
+                        help="Configuration INI file to retrieve database connection settings (default: %(default)s).")
+    parser.add_argument("--key-file",
+                        help="Location to write key file to. Default is to use the first file listed in the "
```
fmigneault: "*Location where to write key file*" seems more readable

fmigneault (changes requested):
I think the initial interaction schemas between Node A/B to establish the trusted register/verify actions between each other is needed. It is still somewhat unclear which request/data flow occurs in that step. With the current schemas, it looks like they can both (or anyone else) simply request their public token without limitation. Therefore, we do not see how the trust between them is established (using only schemas, I know it is described in *Managing the Network* section).

Using only the current schemas, it looks like a bad actor could simply spin up their own node with random public tokens for "validation" with other legit nodes, and potentially receive leaked auth tokens from users of the originating node requesting the resources. The nodes would have to actually invoke that bad-actor node to leak the details, but that might not be that far-fetched if the invocations are embedded within notebooks running inside a given pre-authenticated user kernel. 

There should be something that highlights it is not just any node, but a pre-established trusted link between them, and clarify how all of this happens to guarantee the integrity of the network.

.github/workflows/tests.yml
mishaschwartz: > Is this for non-remote network unit tests?

Yes

> Remote network tests does not need it as well?

It does but in that case it is created automatically when the app starts by setting the `MAGPIE_NETWORK_CREATE_MISSING_PEM_FILE` variable.

> Are non-remote network tests expected to run along other unit tests?

Yes

> What if network mode is disabled for that instance?

For local tests, network mode will be enabled for that specific test case.
For remote tests, the test will be skipped.

See:

https://github.com/Ouranosinc/Magpie/blob/fa4f9e8bc073c15220ad493176f22a03ddc31cf7/tests/utils.py#L743-L760

> They will automatically fail?

No

docs/_static/client-request-flow.drawio
fmigneault: Indeed, I was referring to the new schema in the PR but didn't have anywhere to tag it during review. The questions are relevant to the last diagram of https://github.com/Ouranosinc/Magpie/pull/589#issuecomment-3366447808

> I don't think that mentioning protected resources in this context is a new concept.

From the Magpie POV alone, validating protected resources is simply a matter of invoking the *Effective Permissions* to resolve concrete allowed/denied access by the authenticated user. The specific endpoint to do this in Magpie is using `/users/{id}/resources/{id}/permissions?effective=true` (see https://pavics-magpie--589.org.readthedocs.build/en/589/permissions.html#distinguishing-applied-inherited-and-effective-permissions). With Twitcher, this is done internally via the [`MagpieAdapter`](https://pavics-magpie--589.org.readthedocs.build/en/589/usage.html#docker-application), but this is a technicality to avoid at this point when describing the *Network Mode* behaviour. Therefore, the concept of other protected resources from a different URI at that point is fairly abstract, so it cannot be easily understood how `/some/proteceted/resource` involves Magpie.


> Would saying "Prior to these steps, the user is not authenticated" 

Yes. It is just a matter of timing in the schema since proving the header causes the authentication to happen.

tests/interfaces.py
```diff
@@ -99,6 +103,15 @@ class ConfigTestCase(object):
     test_admin = None                       # type: Optional[Str]
     test_user_name = None                   # type: Optional[Str]  # also used as password when creating test user
     test_group_name = None                  # type: Optional[Str]
+    test_node_host = None                    # type: Optional[Str]
+    test_node_port = None                   # type: Optional[int]
```
fmigneault: misaligned comment

.github/workflows/tests.yml
fmigneault: OK. I just want to double-check my understanding of the test suite. 

Since the new test methods are applied with the `@utils.check_network_mode` decorator onto existing `unittest.TestCase`, and that those classes' setup invoke `cls.setup_network_attrs()`, they are all running with *Network Mode* enabled? Shouldn't they fail in this case instead if the network was not configured correctly? That would validate that configuration setup also works, not silently ignored.

Also, doesn't that mean that nothing tests these API calls with *Network Mode* disabled anymore? 

I would expect to have distinct derived unittest classes that evaluate with/without *Network Mode*. This is similar to how the various https://github.com/Ouranosinc/Magpie/blob/master/tests/interfaces.py classes are reused to validate different branches in the code depending on enabled features or not.

magpie/api/requests.py
mishaschwartz: Why transfer them from weaver? 

Can't I just do:

```py
from magpie.typedefs import AnyAccessPrincipalType, AnyRequestType
```

and then change it to:

```
    # type: (AnyRequestType) -> AnyRequestType
```

fmigneault: In this case, inputs/outputs are not the requests themselves, but a view callable taking a request (and optionally a context), returning a modified one. I have developed this more on Weaver end, and haven't transferred them back in Magpie. (by transfer, I mean copy-paste literally, and then do `from magpie.typedefs import AnyViewCallable`).

magpie/api/management/network/network_views.py
```diff
@@ -0,0 +1,91 @@
...
+@view_config(route_name=s.NetworkTokensAPI.name, request_method="DELETE", decorator=check_network_mode_enabled)
+def delete_network_tokens_view(request):
+    if asbool(get_multiformat_body(request, "expired_only", default=False)):
+        deleted = models.NetworkToken.delete_expired(request.db)
+    else:
+        deleted = request.db.query(NetworkToken).delete()
+    # clean up unused records in the database (no need to keep records associated with anonymous network users)
+    (request.db.query(models.NetworkRemoteUser)
+     .filter(models.NetworkRemoteUser.user_id == None)  # noqa: E711 # pylint: disable=singleton-comparison
+     .filter(models.NetworkRemoteUser.network_token_id == None)  # noqa: E711 # pylint: disable=singleton-comparison
```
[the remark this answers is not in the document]
mishaschwartz: Yeah. sqlalchemy doesn't translate `is None` to a query string properly. It translates `== None` though (`COL_NAME IS NULL`).

But pylint doesn't know that we're doing sqlalchemy stuff so it things we're just comparing a regular python variable to None with `==`.

magpie/cli/create_private_key.py
fmigneault: 😆 yes, maybe this is a French perspective

It is the double use of `to` that sounds odd to me (though it is valid syntax). 
"*Location where the key file must be saved to*" is OK as well. 

Not a big deal though.

magpie/api/management/network/network_views.py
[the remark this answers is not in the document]
fmigneault: OK thanks for validating 👍

docs/configuration.rst
mishaschwartz: > Should there be a Network Group term as well?

There could be. We're currently not referring to it anywhere else in the documentation but we could put it in for the future.

> What is its implication/usage

It's a group that gets created for each network node. Each user who logs in from that network node is part of the group. So if you want to have different access permissions for people from node A vs node B you'd set those permissions on the appropriate network group.

>  could it be needed to have different onces per node?

No, only one per node

mishaschwartz: Actually thinking about it again, there's no real need for a definition since it's not functionally different from a regular group and I think it would be confusing to give it a special definition.

magpie/api/requests.py
mishaschwartz: OK I started doing that but then realized that I was bringing over half of that file since there are so many types that reference each other. I'm not confident I can do it properly so I'll leave a note to update this annotation once those type definitions are ported over in a different PR.

fmigneault: Ok, no problem.

PR #590, opened 2023-10-03
allow service registration forced update of type field

fmigneault:
relates to https://github.com/bird-house/birdhouse-deploy/pull/348

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,7 +11,9 @@ Changes
 
 Features / Changes
 ~~~~~~~~~~~~~~~~~~~~~
-* n/a
+* Allow startup `Service` registration to update the ``type`` field if a change is detected when loading ``providers``
+  configurations. Note that modifying ``type`` of a `Service` with existing `Resource` or `Permission` definitions on
+  the `Service` itself or any of its children could break the `Magpie` instance if those definitions are not compatible.
 
 Bug Fixes
 ~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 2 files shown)


mishaschwartz (approved):
This looks good to me. 

I don't think this needs to be addressed in this PR but I'm a little concerned that this can "break the `Magpie` instance if those definitions are not compatible". 

I did a little experimenting and it looks like an incompatible resource definition will cause a 500 error for certain routes but won't completely break the whole Magpie instance which is good. 

Maybe we should add a line about how to recover in case a breakage occurs. Something like:

- If updating the service type results in unexpected behaviour due to incompatible `Resource` or `Permission` definitions, please manually delete  or update the problematic definitions from the `Magpie` database.

I know it's not a very user-friendly solution but it at least gives the user a recommended course of action.

fmigneault:
@mishaschwartz 

> I don't think this needs to be addressed in this PR but I'm a little concerned that this can "break the `Magpie` instance if those definitions are not compatible".
> 
> I did a little experimenting and it looks like an incompatible resource definition will cause a 500 error for certain routes but won't completely break the whole Magpie instance which is good.

Indeed. This is what I meant by "break". The code assumes that valid combinations of services/resources/permissions where pre-validated by the API/UI/registration script beforehand. If items are combined in a way that is invalid, it won't revalidate everything each time, because that would slow down way too much the execution. 

This is mostly why the `svc.type = svc_type` was not applied before. 
I'm willing to let it be updated for the `force_update` case, but it is a "due at your own risk" kind of operation. 
In most cases, this will break, but for situations such as merging the individual parts of `geoserver[wfs|wms|wps|api]` into `geoserver`, that would work (but not the other way around!). 

> Maybe we should add a line about how to recover in case a breakage occurs.

For this, I think it is more a case of backing up your DB, try it, and revert as needed.

PR #593, opened 2023-10-24
Update CI PR labeler version to fix HTTP error with GitHub token

fmigneault:


fmigneault:
⬆️ Above `ci` tag automatically applied shows that action works as expected.

PR #594, opened 2023-10-25
allow case sensitive regex matches for the `MAGPIE_USER_NAME_EXTRA_REGEX` setting

mishaschwartz:
Ensures that the settings/environment variable ``MAGPIE_USER_NAME_EXTRA_REGEX`` is case sensitive.

Previously, the check was case insensitive meaning that it could not be used to restrict usernames based on case.
For example, setting this value to `^[a-z]+$` would have permit the username `"someuser"` as well as `"Someuser"`. 

Now, the same regular expression will not match `"Someuser"` since case sensitivity in enforced.

### DIFF
--- a/tests/interfaces.py
+++ b/tests/interfaces.py
@@ -7538,6 +7538,27 @@ class Interface_MagpieUI_AdminAuth(AdminTestCase, BaseTestCase):
             utils.check_val_is_in(msg, html.unescape(body))
 
     @runner.MAGPIE_TEST_USERS
+    def test_AddUser_FormSubmit_WithExtraUsernameRegex_CaseInvalid(self):
+        """
+        Check that the extra_user_name_regex setting is used to validate a new user name when the user name is
+        invalid according to that regex because the case is incorrect but is valid according to the ax.PARAM_REGEX.
+
+        .. versionadded:: 3.37
+        """
+        utils.warn_version(self, "extra username regex added", "3.37", skip=True)
+        with utils.mocked_get_settings(settings={"magpie.user_name_extra_regex": "^[a-z]+$"}):
+            data = {"user_name": "UpperCaseUserName", "group_name": get_constant("MAGPIE_USERS_GROUP"),
+                    "email": "{}@mail.com".format(self.test_user_name),
+                    "password": self.test_user_name, "confirm": self.test_user_name}
+            path = "/ui/users/add"
+            form = "add_user_form"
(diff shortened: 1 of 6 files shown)


tests/test_magpie_api.py
```diff
@@ -489,6 +489,25 @@ def test_PostUsers_WithExtraRegex_InvalidExtraRegex(self):
...
 
+    @runner.MAGPIE_TEST_USERS
+    def test_PostUsers_WithExtraRegex_CaseInvalidExtraRegex(self):
+        """
+        Check that the user_name_extra_regex setting is used to validate a new user name when the user name is
+        invalid according to that regex because the case is incorrect but is valid according to the ax.PARAM_REGEX.
+
+        .. versionadded:: 3.37
+        """
+        utils.warn_version(self, "extra username regex added", "3.37", skip=True)
```
fmigneault, line 500: 3.37.1 / versionchanged

tests/interfaces.py
```diff
@@ -7537,6 +7537,27 @@ def test_AddUser_FormSubmit_WithExtraUsernameRegex_Invalid(self):
...
 
+    @runner.MAGPIE_TEST_USERS
+    def test_AddUser_FormSubmit_WithExtraUsernameRegex_CaseInvalid(self):
+        """
+        Check that the extra_user_name_regex setting is used to validate a new user name when the user name is
+        invalid according to that regex because the case is incorrect but is valid according to the ax.PARAM_REGEX.
+
+        .. versionadded:: 3.37
+        """
+        utils.warn_version(self, "extra username regex added", "3.37", skip=True)
```
fmigneault, line 7548: 3.37.1

I think `versionchanged` is better here.

fmigneault (changes requested):
`param_compare=re.compile(...)` can be passed directly with any desired flags.
No need for the `regex_flags` parameter.

I like the idea of `re.X` being the default however, and not assume `re.I` by default.
All current regex used `[A-Za-z]` explicitly anyway, so they should still work with only `re.X`.

fmigneault (approved):
Looks good. Just waiting on tests to complete validation.

mishaschwartz:
@fmigneault  The tests are failing because `data.ceda.ac.uk` is unresponsive right now: https://github.com/Ouranosinc/Magpie/actions/runs/6655721960/job/18098444930?pr=594

fmigneault:
Validated that CEDA has planned down server until November 7.
Validated that tests and other checks than `links` run successfully.

PR #596, opened 2023-11-08
do not allow conflicting usernames where only difference is case

### LINKED ISSUE #595: [BUG] login confuses usernames that differ in terms of upper/lower case
**Describe the bug**

Ziggurat foundation converts usernames to lowercase before looking them up in the database: https://github.com/ergo/ziggurat_foundations/blob/0.9.1/ziggurat_foundations/models/services/user.py#L326-L328

This means:

- if a user named "Test" exists, logging in with any of "test", "TEST", "tEst", etc. will all log in as the "Test" user
- if users named "test" and "Test" exist, both users will be logged in as the "test" user ("Test" will never be able to log in)

Prior to this change https://github.com/Ouranosinc/Magpie/pull/594/files#diff-c42ba4dc908c0ec2a09c7c0e7ff3864d8be94c1e0cbdb6e29655f62aa29d7153R252 Magpie would not allow users to have usernames that differ in terms of case only. We should probably go back to that.

**To Reproduce**

1. Create a user "Test"
2. Log in with the username "test" (result: login success as the user "Test")
3. Create a user "test" with a different password than "Test"
4. Log in with the username "Test" (result: login failure)

**Expected behavior**

Users whose usernames differ in terms of capitalization only, should be treated as distinct users in all parts of the application, including logins.

mishaschwartz:
Do not allow users to have the same username that only differs in terms of upper/lowercase
Ziggurat foundations assumes that users will not have usernames that differ in terms of case.
This means that if the database contains a user with the username "Test" and another with the username "test". The login procedure will not know to differentiate the two.
 
Resolves #595

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,7 +9,11 @@ Changes
 `Unreleased <https://github.com/Ouranosinc/Magpie/tree/master>`_ (latest)
 ------------------------------------------------------------------------------------
 
-* Nothing new for the moment.
+* Do not allow users to have the same username that only differs in terms of upper/lowercase
+  Ziggurat foundations assumes that users will not have usernames that differ in terms of case.
+  This means that if the database contains a user with the username "Test" and another with the username "test". The
+  login procedure will not know to differentiate the two.
+  (fixes `#595 <https://github.com/Ouranosinc/Magpie/issues/595>`_).
 
 .. _changes_3.37.1:
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,10 +9,13 @@ Changes
 `Unreleased <https://github.com/Ouranosinc/Magpie/tree/master>`_ (latest)
 ------------------------------------------------------------------------------------
 
-* Do not allow users to have the same username that only differs in terms of upper/lowercase
-  Ziggurat foundations assumes that users will not have usernames that differ in terms of case.
-  This means that if the database contains a user with the username "Test" and another with the username "test". The
-  login procedure will not know to differentiate the two.
+Bug Fixes
+~~~~~~~~~
+
+* Do not allow any `User` to have the same ``user_name`` as another `User` that only differs in terms of upper/lowercase
+  Ziggurat foundations assumes that a `User` will not have a ``user_name`` that differs from another only in terms of case.
+  This means that if the database contains a `User` with the ``user_name`` "Test" and another with the ``user_name``
+  "test". The login procedure will not know to differentiate the two.
   (fixes `#595 <https://github.com/Ouranosinc/Magpie/issues/595>`_).
 
 .. _changes_3.37.1:
--- a/magpie/api/management/user/user_utils.py
+++ b/magpie/api/management/user/user_utils.py
@@ -135,7 +135,7 @@ def create_user(user_name,              # type: Str
     user_params = {"with_param": False}
     if user_checked:
         user_params["with_param"] = True
-        if compare_digest(user_checked.user_name, user_name):
+        if compare_digest(user_checked.user_name.lower(), user_name.lower()):
             user_params["msg_on_fail"] = s.User_Check_Name_ConflictResponseSchema.description
             user_params["param_name"] = "user_name"
             user_params["param_content"] = {"value": user_name, "conditions": [{"unique": False}]}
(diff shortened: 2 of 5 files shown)


magpie/api/management/user/user_utils.py
```diff
@@ -135,7 +135,7 @@ def _get_group(grp_name):
     user_params = {"with_param": False}
     if user_checked:
         user_params["with_param"] = True
-        if compare_digest(user_checked.user_name, user_name):
+        if compare_digest(user_checked.user_name.lower(), user_name.lower()):
```
fmigneault, line 138: We must avoid tweaking the values. That can give timing attack information. If lowercase is enforced anyway from the next version, that will not be necessary.

magpie/ui/utils.py
```diff
@@ -567,7 +567,7 @@ def create_user(self, data):
         if len(user_name) > get_constant("MAGPIE_USER_NAME_MAX_LENGTH", self.request):
             data["invalid_user_name"] = True
             data["reason_user_name"] = "Too Long"
-        if user_name in [usr["user_name"] for usr in user_details]:
+        if user_name.lower() in [usr["user_name"].lower() for usr in user_details]:
```
fmigneault, line 570: Not needed here. Enforce it directly from the DB/API.

CHANGES.rst
```diff
@@ -9,7 +9,11 @@ Changes
 `Unreleased <https://github.com/Ouranosinc/Magpie/tree/master>`_ (latest)
 ------------------------------------------------------------------------------------
 
-* Nothing new for the moment.
+* Do not allow users to have the same username that only differs in terms of upper/lowercase
+  Ziggurat foundations assumes that users will not have usernames that differ in terms of case.
+  This means that if the database contains a user with the username "Test" and another with the username "test". The
+  login procedure will not know to differentiate the two.
+  (fixes `#595 <https://github.com/Ouranosinc/Magpie/issues/595>`_).
```
fmigneault, line 16: Use format ``` `User` ``` and ``` ``user_name`` ``` as employed for the rest of the docs.

Missing dot at the end of the first line.
Move the `(fixes ...)` definition after the first line (after upper/lowercase), no dot in between. 

Move under section 
```rst
Bug Fixes
~~~~~~~~~~~~~~~~~~~~~
```

fmigneault (changes requested):
The valid user name regex should be updated to directly disallow uppercase.
We shouldn't have to go back everywhere to apply `lower()`.

There should also be a new migration script (see the makefile migrate/revision target to generate the next script) that applies `lower()` to any existing user name just to make sure.

Finally, I propose that another more explicit message is added to `check_user` utility to verify if `user_name != user_name.lower()` and raise the relevant error that will better guide the registering user why their user name is considered invalid.

fmigneault:
@tlvu 
Just FYI to validate on your side if this is already applied in your platforms.
The (eventual) migration script could break deployment if there is a conflicting case-insensitive user name. In the meantime, if there are such users, they probably get funky responses anyway.

tlvu:
> @tlvu Just FYI to validate on your side if this is already applied in your platforms. The (eventual) migration script could break deployment if there is a conflicting case-insensitive user name. In the meantime, if there are such users, they probably get funky responses anyway.

@fmigneault Thanks for the heads up.  Due to the sanitized `{username}` from JupyterHub, we do not have duplicate capital case in our usernames.

mishaschwartz:
> The valid user name regex should be updated to directly disallow uppercase.

I'm very confused now. The whole point of the strategy that you suggested here: https://github.com/bird-house/birdhouse-deploy/pull/396 (especially overriding the `normalize_username` function) was to allow there to be usernames with uppercase characters. 

If we now want to disallow all uppercase characters here, we could have avoided making those changes in the first place.

> We shouldn't have to go back everywhere to apply lower().

This is what ziggurat foundations does, if we want to conform with that code then we have to ensure that we're matching names in a case-insensitive way.

magpie/api/management/user/user_utils.py
mishaschwartz: see https://github.com/Ouranosinc/Magpie/pull/596#issuecomment-1804221294

magpie/ui/utils.py
mishaschwartz: What is the point of these checks then?

fmigneault: The best would be to avoid duplicating the logic in the UI and leave all validations on the API side that can report more specific error messages. 
However, given how small this change is, it is fine to leave it as is.
Disregard my previous comment.

mishaschwartz:
Closed in favour of #597

PR #597, opened 2023-11-11
Ensure that ``user_name`` values for all `User` are lowercase and do not contain whitespace

### LINKED ISSUE #595: [BUG] login confuses usernames that differ in terms of upper/lower case

mishaschwartz:
Ziggurat foundations assumes that a `User` will not have a ``user_name`` that differs from another only in terms of case. The simplest way to enforce this is to ensure that all ``user_name`` values are lowercase. 

Previously, this was not enforced so we could create two `User` which could not be differentiated properly.

This change includes a database migration that will convert all ``user_name`` that contain uppercase characters to lowercase. This may cause a database conflict if there are two ``user_name`` values that differ only in terms of case. For example "Test" and "test". If this occurs, please manually update those ``user_name`` values to no longer conflict and try the migration again.

This also prevents new users from being created that contain whitespace.

Replaces #596 
Resolves #595

### DIFF
--- a/magpie/alembic/versions/2023-11-10_91af68bcdc67_case_insensitive_user_name_constraint.py
+++ b/magpie/alembic/versions/2023-11-10_91af68bcdc67_case_insensitive_user_name_constraint.py
@@ -38,7 +38,7 @@ def upgrade():
     except sa.exc.IntegrityError as e:
         raise Exception("{}\nPlease manually update conflicting user_names and try again".format(e)) from e
     session = Session(bind=op.get_bind())
-    for user in session.execute(sa.select(users)):
+    for user in session.execute(sa.select([users])):
         lower_user_name = user.user_name.lower()
         if user.user_name != lower_user_name:
             print(
--- a/magpie/alembic/versions/2023-11-10_91af68bcdc67_case_insensitive_user_name_constraint.py
+++ b/magpie/alembic/versions/2023-11-10_91af68bcdc67_case_insensitive_user_name_constraint.py
@@ -5,9 +5,10 @@ Revision ID: 91af68bcdc67
 Revises: 5e5acc33adce
 Create Date: 2023-11-10 15:58:23.068465
 """
-
+import sqlalchemy
 import sqlalchemy as sa
 from alembic import op
+from sqlalchemy import func
 from sqlalchemy.orm.session import sessionmaker
 
 # Revision identifiers, used by Alembic.
@@ -27,15 +28,19 @@ users = sa.table(
 
 
 def upgrade():
-    op.create_index(
-        "ix_users_user_name_unique_case_insensitive",
-        "users",
-        [sa.text("lower(user_name)")],
-        unique=True,
-    )
+    # If users with conflicting user_names already exist, the following will fail. The conflicting user_names
+    # must be updated manually and then this migration script can be re-run.
+    try:
+        op.create_index(
+            "ix_users_user_name_unique_case_insensitive",
+            "users",
+            [sa.text("lower(user_name)")],
+            unique=True,
+        )
+    except sqlalchemy.exc.IntegrityError as e:
+        raise Exception("{}\nPlease manually update conflicting user_names and try again".format(e)) from e
     session = Session(bind=op.get_bind())
-    for user in session.execute(sa.select(users)):
-        session.execute(users.update().where(users.c.id == user.id).values(user_name=user.user_name.lower()))
+    session.execute(users.update().values({users.c.user_name: func.lower(users.c.user_name)}))
     session.commit()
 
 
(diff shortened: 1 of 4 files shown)


magpie/alembic/versions/2023-11-10_91af68bcdc67_case_insensitive_user_name_constraint.py
```diff
@@ -0,0 +1,44 @@
...
+def upgrade():
+    op.create_index(
+        "ix_users_user_name_unique_case_insensitive",
+        "users",
+        [sa.text("lower(user_name)")],
+        unique=True,
+    )
+    session = Session(bind=op.get_bind())
+    for user in session.execute(sa.select(users)):
+        session.execute(users.update().where(users.c.id == user.id).values(user_name=user.user_name.lower()))
```
fmigneault, line 38: I think this could be done with a single query: 
```python
session.execute(users.update().values({users.c.id: func.lower(users.c.id)})
```

fmigneault (dismissed):
Looks good. Only minor potential patch, but not blocking.

tlvu:
Oh, I only noticed this PR now.  Great !  More harmonization with JupyterHub `{username}`.

Question about the DB migration, if it encounters existing username with uppercase and try to convert to a lowercase version, will it display which users were converted so the node admin can notify the users impacted they have to now use lowercase to login?

What if the conversion clashes with existing lowercase version?

fmigneault:
> will it display which users were converted so the node admin can notify the users impacted they have to now use lowercase to login?
It won't display it. That could be a good reason to keep the `for user in session.execute(sa.select(users)):` approach in the migration script. A `LOGGER.warning` could be employed in the loop to clearly indicate it to the admin within `pavics-compose logs -f magpie`. 

> What if the conversion clashes with existing lowercase version?

The migration will purposely crash. Magpie will not be able to boot, and therefore the rest of the stack also. 
It would be up to the admin to investigate and decide how they want to remedy the situation (inform the user of a rename, delete, etc.).

tlvu:
> > What if the conversion clashes with existing lowercase version?
> 
> The migration will purposely crash. Magpie will not be able to boot, and therefore the rest of the stack also. It would be up to the admin to investigate and decide how they want to remedy the situation (inform the user of a rename, delete, etc.).

I assume when Magpie crashes, it will display/log the conflicting existing username causing the crash and the matching uppercase version?

fmigneault:
> I assume when Magpie crashes, it will display/log the conflicting existing username causing the crash and the matching uppercase version?

Magpie will report the error to boot after the configured number of DB connect/migration retries (default 5).
I assume `sqlalchemy` will report some kind of error being unable to update the value due to the "no duplicate" index rule.
However, the migration script itself could provide a log call to be more explicit about the known potential issue and how to address it.

magpie/alembic/versions/2023-11-10_91af68bcdc67_case_insensitive_user_name_constraint.py
mishaschwartz: I'm not sure I understand this query... 

Did you mean: 

```python
session.execute(users.update().values({users.c.user_name: func.lower(users.c.user_name)}))
```

That seems to work. I'll update it

tlvu:
> Since the case insensitive unique index is created first (before existing usernames are converted to lowercase). Alembic will report an error indicating which values are in conflict before any conversion happens.
> 
> This will give the user a good idea of which names are in conflict and they can update the values manually. I've added a comment in the migration script as well with this information and added additional information to the error message.


@mishaschwartz Thanks for logging the error when migration fails.

> > will it display which users were converted so the node admin can notify the users impacted they have to now use lowercase to login?
> 
> It won't display it (in the current state). That could be a good reason to keep the `for user in session.execute(sa.select(users)):` approach in the migration script. A `LOGGER.warning` could be employed in the loop to clearly indicate it to the admin within `pavics-compose logs -f magpie`. Warning would be used only if `user.user_name != user.user_name.lower()` to avoid being unnecessarily noisy.



When migration succeeds, please also log which users were migrated so the node admin can notify those users to use lowercase instead of uppercase.

fmigneault (approved):

PR #606, opened 2024-04-25
Maintenance update

fmigneault:
## Changes

* Add support of Python 3.12.
  
> [!NOTE]  
> Requires patch of ``authomatic.six.moves`` in ``magpie.__init__.py`` to work around vendor-specific definitions.
    Fix inspiration from https://github.com/dpkp/kafka-python/issues/2401#issuecomment-1760208950.
    Pending official fix (see https://github.com/authomatic/authomatic/issues/233).

* Drop support of Python 3.5, 3.6 and 3.7. Minimum version is Python 3.8.
* Pin ``gunicorn>=22`` to address CVE-2024-1135.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -48,7 +48,7 @@ jobs:
       matrix:
         # os: [ubuntu-latest, windows-latest]
         os: [ubuntu-latest]
-        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11"]
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
         allow-failure: [false]
         test-case: [test-local]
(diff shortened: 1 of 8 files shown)


fmigneault:
>[!NOTE]
> Failing `tests (ubuntu-latest, 3.7, false, test-local)` is expected, since it was removed from the CI. 
> The rule will be removed from the PR merge requirements once integrate (such that any PR in the meantime still respect the requirement). 
> Coverage diff is only related to the Python 3.12 patch, which is not worth testing by itself (the fact that all other tests executed with 3.12 is enough of an indicator that it works).

mishaschwartz (approved):
This looks great and I am 100% in favour of dropping support for python < 3.8 as well!

If this gets merged, I'll go and update #589 since there's a bunch of extra code in there to support versions < 3.8.

PR #634, opened 2024-12-09
Allow `metadata_type.prefixes` and `data_type.prefixes` in `ServiceTHREDDS` configuration to contain `/` character

### LINKED ISSUE #633: [Feature] Allow `metadata_type.prefixes` and `data_type.prefixes` in `ServiceTHREDDS` configuration to contain `/` character
**Is your feature request related to a problem? Please describe.**

THREDDS can have service paths that include subpaths (ex: `/ncss/grid` and `ncss/point`). We should be able to define a configuration file for the Magpie THREDDS service like so:

```yaml
...
data_type:
  prefixes:
    ...
    - ncss
    - ncss/grid
    - ncss/point
...
```

So that part of the URL path that defines the service can be treated as a prefix.

Currently, we cannot include `/` in prefix values because Magpie splits the URL path on `/` characters.

The current workaround is to just include the top-most path part as the prefix and apply permissions to the rest of the path parts as if they were directories. However, this means that we need to define permissions multiple times for the same resource. 
For example:

- Right now we have URL paths like: 

  - `${FQDN}/twitcher/ows/proxy/thredds/ncss/datasets/...`
  - `${FQDN}/twitcher/ows/proxy/thredds/dodsC/datasets/...`

  So if we want to set specific permissions on the `datasets/` directory we can do that by setting one directory permission rule on the `datasets/` subdirectory and it will work for all services.

  But if we instead have:

    - `${FQDN}/twitcher/ows/proxy/thredds/ncss/point/datasets/...`
    - `${FQDN}/twitcher/ows/proxy/thredds/dodsC/datasets/...`

  Then we would need to set the same permission rule on `datasets/` as well as `point/datasets/`. If we don't set the same rule on `point/datasets/` then the rule won't apply to resources accessed by the `ncss/point` service.

**Describe the solution you'd like**

Parse the prefix values differently so that values that include `/` can be matched. I suggest that [`permission_requested`](https://github.com/Ouranosinc/Magpie/blob/dc0a9a9871f13f1c55b219d644245e9cc3e1c96e/magpie/services.py#L1432) function should detect that a prefix includes a `/` and applies multiple matches to each path part in order.

**Additional context**

See initial discussion of this issue here: https://github.com/bird-house/birdhouse-deploy/pull/413#discussion_r1852517144

mishaschwartz:
Resolves #633 

Also updates a broken docs link (unrelated to this PR but required to make CI tests run)

### DIFF
--- a/magpie/services.py
+++ b/magpie/services.py
@@ -1440,7 +1440,7 @@ class ServiceTHREDDS(ServiceInterface):
             for pattern_prefix in prefixes:  # type: Str
                 if not path_parts and pattern_prefix is None:
                     return permission
-                elif pattern_prefix is not None:
+                if pattern_prefix is not None:
                     pattern_prefix = pattern_prefix.strip("/")
                     path_prefix = "/".join(path_parts[:pattern_prefix.count("/") + 1])
                     if self.is_match(path_prefix, pattern_prefix) is not None:
--- a/magpie/services.py
+++ b/magpie/services.py
@@ -1433,18 +1433,18 @@ class ServiceTHREDDS(ServiceInterface):
         # type: () -> PermissionRequested
         cfg = self.get_config()
         path_parts = self.get_path_parts()
-        path_prefix = None  # in case of no `<prefix>`, simulate as `null`
-        if path_parts:
-            path_prefix = path_parts[0]
         for prefixes, permission in [
             (cfg["metadata_type"]["prefixes"], Permission.BROWSE),  # first to favor BROWSE over READ prefix conflicts
             (cfg["data_type"]["prefixes"], Permission.READ),
         ]:
             for pattern_prefix in prefixes:  # type: Str
-                if path_prefix is None and pattern_prefix is None:
+                if not path_parts and pattern_prefix is None:
                     return permission
-                if self.is_match(path_prefix, pattern_prefix) is not None:
-                    return permission
+                elif pattern_prefix is not None:
+                    pattern_prefix = pattern_prefix.strip("/")
+                    path_prefix = "/".join(path_parts[:pattern_prefix.count("/")+1])
+                    if self.is_match(path_prefix, pattern_prefix) is not None:
+                        return permission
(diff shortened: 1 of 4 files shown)


magpie/services.py
```diff
@@ -1382,6 +1382,20 @@ def get_path_parts(self):
                 return path_parts[1:]  # remove extra '' added by split
         return path_parts
 
+    def _get_path_prefix_permission(self, path_parts):
+        # type: (List[str]) -> Tuple[Optional[Str], PermissionRequested]
+        cfg = self.get_config()
+        for prefixes, permission in ((cfg["metadata_type"]["prefixes"], Permission.BROWSE),
+                                     (cfg["data_type"]["prefixes"], Permission.READ)):
+            for prefix in prefixes:
```
fmigneault, line 1390: Please use `pattern_prefix` here as in the original code. I think it is important to get the hint that we are dealing with potential patterns here, and not just plain strings, hence why some matching operations are required rather than plain compares.

tests/test_services.py
```diff
@@ -580,7 +581,8 @@ def test_ServiceTHREDDS_custom_config(self):
             utils.check_raises(lambda: self.ows.check_request(req), OWSAccessForbidden, msg=msg)
 
         # using unknown prefixes, otherwise allowed file should always be denied
-        unknown_prefixes = ["ncml", "dap4"]  # purposely take normally allowed THREDDS prefixes, validate active config
+        # purposely take normally allowed THREDDS prefixes, validate active config
+        unknown_prefixes = ["ncml", "dap4", "ncss/point"]
```
fmigneault, line 585: Can you add 2 extra `ncss` and `grid` on their own, just to make sure the prefix matching doesn't incorrectly do any partial match of the pre-split parts?

fmigneault (commented):
Looks good. Just minor comments to address.

fmigneault (approved):
All good. Just waiting for CI to validate.

PR #650, opened 2026-04-21
fix incorrect reserved user names validations

fmigneault:
# Description

Original message erroneously employed the `"email"` message, which made debugging of Cowbird response hard to interpret. It now employs the expected `"user_name + reserved keyword"` message.

While debugging specifically an "admin" use case, it was ambiguous to detect the cause because of the errorneous parameter names.

## Changes

* Fix incorrect message employed in error response when validating reserved user names.
* Replace ``magpie.api.management.user.user_utils.check_user_info`` parameter ``check_not_anonymous``  to ``check_not_reserved`` to better reflect that it valides reserved user names and emails for both   the ``MAGPIE_ANONYMOUS_USER`` and ``MAGPIE_ADMIN_USER`` special users.

PR #651, opened 2026-04-21
ignore expected protected user/group create on permission registration

fmigneault:
## Description

Fix self-induced error on reserved keywords (special `Users`) when resolving `Permission` registrations.
  Notably, when a `Webhook` request is provided with a `Permission` batch to generate or when loading registration configurations, both methods employ the ``magpie.register`` functions. These perform an automatic creation of `User`/`Group`/`Permission` definitions as needed to make a seamless integration of referenced resources. However, if a reserved `User` is referenced for a `Permission` creation on it, while that creation is itself valid, the auto-creation procedure of the `User` would yield a distinct HTTP error that was not handled properly by the registration process.

## References

- based on #650 since the message fix is needed to handle it
- relates to https://github.com/Ouranosinc/cowbird/pull/103


mishaschwartz (changes requested):
Looks good! 

Please add a test for the new case where the permission is created for a reserved user.

fmigneault:
Test added.

mishaschwartz (dismissed):

PR #656, opened 2026-05-22
verify hooks

mishaschwartz:
Add new hook type which is called before verify and can be used to force the verify endpoint to return a "403" response.

Must be implemented with https://github.com/bird-house/twitcher/pull/151

TODO:

- [ ] add tests
- [ ] update documentation
- [ ] update changelog

### DIFF
--- a/magpie/adapter/__init__.py
+++ b/magpie/adapter/__init__.py
@@ -1,4 +1,5 @@
 import copy
+import functools
 import inspect
 import re
 import warnings
@@ -116,10 +117,13 @@ if TYPE_CHECKING:
         JSON,
         AnyResponseType,
         AnySettingsContainer,
+        Callable,
+        Iterable,
         ServiceConfigItem,
         ServiceHookConfigItem,
         ServiceHookType,
-        Str
+        Str,
+        Tuple,
     )
 
     from twitcher.models.service import ServiceConfig  # noqa  # pylint: disable=E0611  # Twitcher >= 0.6.3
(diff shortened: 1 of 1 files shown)


magpie/adapter/__init__.py
```diff
@@ -302,11 +306,10 @@ def configurator_factory(self, container):  # noqa: R0201
...
         return config
 
-    def _apply_hooks(self, instance, service_name, hook_type, method, path, query):
-        # type: (Union[Request, Response], Str, ServiceHookType, Str, Str, Str) -> Union[Request, Response]
-        """
-        Executes the hooks processing chain.
-        """
+    def _discover_hooks(self, instance, service_name, hook_type, method, path, query):
+        # type: (Union[Request, Response], Str, ServiceHookType, Str, Str, Str, bool) -> Iterable[Tuple[Callable, Str, Str]]
+        """Yields hook functions."""
```
fmigneault, line 311: Use the multiline docstring.
Same for others.

magpie/adapter/__init__.py
```diff
@@ -391,12 +414,30 @@ def response_hook(self, response, service):
...
         )
         return response
 
+    def verify_hook(self, request, service):
+        # type: (Request, ServiceConfig) -> bool
+        """
+        Apply hook that can apply additional logic used to verify whether a request should be rejected.
+
+        .. versionadded:: 5.0.3
+            Requires ``Twitcher >= 0.11.2``.
```
fmigneault, line 429: magpie 5.2.0
twitcher 0.12.0

fmigneault (commented):
Will need the corresponding updates when available (from https://github.com/bird-house/twitcher/pull/151): 
https://github.com/Ouranosinc/Magpie/blob/9a35799db2365b06d1a741ea26cb8e6e9d6a5cdd/requirements-dev.txt#L18
https://github.com/Ouranosinc/Magpie/blob/9a35799db2365b06d1a741ea26cb8e6e9d6a5cdd/Dockerfile.adapter#L6

mishaschwartz:
This feature is no longer needed and I'm not convinced that this makes sense really to be able to short-circuit the entire auth mechanism set up by magpie and twitcher.

If anyone is interested in revisiting this later please feel free to re-open this PR.

